PR #55, opened 2019-08-08
implemented unit test cases for SDK

Suryakantj:
implemented unit test cases for SDK

### DIFF
--- a/cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
+++ b/cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
@@ -19,8 +19,8 @@ import static org.mockito.Mockito.mock;
 public class ConsentLibBuilderTest {
 
 
-    ConsentLibBuilder consentLibBuilder;
-    ConsentLib.Callback callback;
+    private ConsentLibBuilder consentLibBuilder;
+    private ConsentLib.Callback callback;
 
     @Before
     public void initConsentLibBuilder() {
@@ -31,6 +31,12 @@ public class ConsentLibBuilderTest {
 
             }
         };
+    }
+
+    public Field getDeclaredFieldAccess(String fieldName) throws Exception{
+        Field member = ConsentLibBuilder.class.getDeclaredField(fieldName);
+        member.setAccessible(true);
+        return member;
     }
 
     @Test(expected = ConsentLibException.class)
@@ -122,25 +128,20 @@ public class ConsentLibBuilderTest {
     public void testSetTargetingParamString() throws Exception {
         String key = "key";
         String stringValue = "stringValue";
-        int intValue = 2;
-        Object objValue = "objValue";
         EncodedParam targetingParamsString;
         consentLibBuilder.setTargetingParam(key, stringValue);
 
-        Field member = ConsentLibBuilder.class.getDeclaredField("targetingParams");
-        member.setAccessible(true);
-
+        Field localMember = getDeclaredFieldAccess("targetingParams");
 
         JSONObject jsonObject = new JSONObject();
         jsonObject.put(key, stringValue);
 
-        assertEquals(jsonObject.toString() , member.get(consentLibBuilder).toString());
+        assertEquals(jsonObject.toString() , localMember.get(consentLibBuilder).toString());
 
         targetingParamsString = new EncodedParam("targetingParams", jsonObject.toString());
         Method method = ConsentLibBuilder.class.getDeclaredMethod("setTargetingParamsString");
         method.setAccessible(true);
         method.invoke(consentLibBuilder);
-        //consentLibBuilder.setTargetingParamsString(targetingParamsString);
         assertEquals(targetingParamsString.toString(), consentLibBuilder.targetingParamsString.toString());
 
     }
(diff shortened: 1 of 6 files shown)


cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+public class SourcePointClientBuilderTest {
+    SourcePointClientBuilder sourcePointClientBuilder;
+
+    @Before
+    public void setUp() throws ConsentLibException {
+        sourcePointClientBuilder = new SourcePointClientBuilder(808, "siteName", true);//mock(SourcePointClientBuilder.class);
```
andresilveirah: What if instead of using an instance variable we turned the initialisation of the `SourcePointClientBuilder` into a function? Example: 
``` java
private SourcePointClientBuilder defaultClientBuilder() {
  return new SourcePointClientBuilder(808, "siteName", true);
}
```
This way we can call it any time in our tests cases, instead of having to use an assignment 
```java
sourcePointClientBuilder.setCmpDomain("cmpDomain");
...
SourcePointClientBuilder sourcePointClientBuilder1 = sourcePointClientBuilder;
```
We'll then:
```java
public void setMmsDomain() throws Exception {
        SourcePointClientBuilder clientBuilder = defaultClientBuilder()
        clientBuilder.setMmsDomain("mmsDomain");
        Field member = SourcePointClientBuilder.class.getDeclaredField("mmsDomain");
        member.setAccessible(true);

        assertEquals("mmsDomain", member.get(clientBuilder).toString());
    }
```

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public void setMmsDomain() throws Exception{
+        sourcePointClientBuilder.setMmsDomain("mmsDomain");
+        Field member = null;
+
+        member = SourcePointClientBuilder.class.getDeclaredField("mmsDomain");
+
+        member.setAccessible(true);
+        SourcePointClientBuilder sourcePointClientBuilder1 = sourcePointClientBuilder;
+
+        assertEquals("mmsDomain", member.get(sourcePointClientBuilder1).toString());
```
andresilveirah: The code to get set a value, turn its member field public and asserting its value is repeated 3 times. Maybe we could refactor it to avoid duplication. If we use my suggestion above, we could write a method like:

```java private Field getPublicField(String fieldName, Object subject) {
  Field field = subject.class.getDeclaredField(fieldName);
  member.setAccessible(true);
  return field;
```

Then, in our tests:
```java
    @Test
    public void setCmpDomain() throws Exception{
        SourcePointClientBuilder clientBuilder = defaultClientBuilder()
        clientBuilder.setMmsDomain("mmsDomain");

        assertEquals("mmsDomain", getPublicField("mmsDomain", clientBuilder).toString());
    }
```

What you think?

Suryakantj: will work on this.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
```diff
@@ -0,0 +1,185 @@
...
+        JSONObject jsonObject = new JSONObject();
+        jsonObject.put(key, stringValue);
+
+        assertEquals(jsonObject.toString() , member.get(consentLibBuilder).toString());
+
+        targetingParamsString = new EncodedParam("targetingParams", jsonObject.toString());
+        Method method = ConsentLibBuilder.class.getDeclaredMethod("setTargetingParamsString");
+        method.setAccessible(true);
+        method.invoke(consentLibBuilder);
+        //consentLibBuilder.setTargetingParamsString(targetingParamsString);
```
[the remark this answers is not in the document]
Suryakantj: I will remove the commented code.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
Suryakantj: will work on this.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
Suryakantj: implemented as suggested

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
Suryakantj: implemented as suggested.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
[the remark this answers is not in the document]
Suryakantj: removed the commented code.

PR #56, opened 2019-08-12
Repeat gdpr status check

Suryakantj:
now gdpr status check in repeated each time SDK is initialized.

### DIFF
--- a/cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
+++ b/cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
@@ -24,13 +24,8 @@ public class ConsentLibBuilderTest {
 
     @Before
     public void initConsentLibBuilder() {
-        consentLibBuilder = new ConsentLibBuilder(808, "cybage.com", mock(Activity.class));
-        callback = new ConsentLib.Callback() {
-            @Override
-            public void run(ConsentLib c) {
-
-            }
-        };
+        consentLibBuilder = new ConsentLibBuilder(123, "example.com", mock(Activity.class));
+        callback = c -> {  };
     }
 
     public Field getDeclaredFieldAccess(String fieldName) throws Exception{
@@ -40,95 +35,94 @@ public class ConsentLibBuilderTest {
     }
 
     @Test(expected = ConsentLibException.class)
-    public void testBuild() throws ConsentLibException {
+    public void build() throws ConsentLibException {
         ConsentLib consentLib = mock(ConsentLib.class);
         assertEquals(consentLib, consentLibBuilder.build());
     }
 
     @Test
-    public void testSetMessageTimeout() {
+    public void setMessageTimeout() {
         consentLibBuilder.setMessageTimeOut(20000);
         assertEquals(20000, consentLibBuilder.defaultMessageTimeOut);
     }
 
     @Test
-    public void testSetPage() {
+    public void setPage() {
         String page = "page";
         consentLibBuilder.setPage(page);
         assertEquals(page, consentLibBuilder.page);
     }
 
     @Test
-    public void testSetViewGroup() {
+    public void setViewGroup() {
         ViewGroup v = mock(ViewGroup.class);
         consentLibBuilder.setViewGroup(v);
         assertEquals(v, consentLibBuilder.viewGroup);
     }
 
     @Test
-    public void testSetOnMessageChoiceSelect() {
+    public void setOnMessageChoiceSelect() {
         consentLibBuilder.setOnMessageChoiceSelect(callback);
         assertEquals(callback, consentLibBuilder.onMessageChoiceSelect);
     }
 
     @Test
-    public void testSetOnInteractionComplete() {
+    public void setOnInteractionComplete() {
         consentLibBuilder.setOnInteractionComplete(callback);
         assertEquals(callback, consentLibBuilder.onInteractionComplete);
     }
 
     @Test
-    public void testSetOnMessageReady() {
+    public void setOnMessageReady() {
         consentLibBuilder.setOnMessageReady(callback);
         assertEquals(callback, consentLibBuilder.onMessageReady);
     }
 
     @Test
-    public void testSetOnErrorOccurred() {
+    public void setOnErrorOccurred() {
         consentLibBuilder.setOnErrorOccurred(callback);
         assertEquals(callback, consentLibBuilder.onErrorOccurred);
     }
 
     @Test
-    public void testSetStage() {
+    public void setStage() {
         boolean stage = true;
         consentLibBuilder.setStage(stage);
         assertEquals(stage, consentLibBuilder.stagingCampaign);
     }
 
     @Test
-    public void testSetInternalStage() {
+    public void setInternalStage() {
         boolean stage = true;
         consentLibBuilder.setInternalStage(stage);
         assertEquals(stage, consentLibBuilder.staging);
     }
 
     @Test
-    public void testSetInAppMessagePageUrl() {
+    public void setInAppMessagePageUrl() {
         String inAppMessageUrl = "inAppMessageUrl";
         consentLibBuilder.setInAppMessagePageUrl(inAppMessageUrl);
         assertEquals(inAppMessageUrl, consentLibBuilder.msgDomain);
     }
 
     @Test
-    public void testSetMmsDomain() {
+    public void setMmsDomain() {
         String mmsDomain = "mmsDomain";
         consentLibBuilder.setMmsDomain(mmsDomain);
         assertEquals(mmsDomain, consentLibBuilder.mmsDomain);
     }
 
     @Test
-    public void testSetCmpDomain() {
+    public void setCmpDomain() {
         String cmpDomain = "cmpDomain";
         consentLibBuilder.setCmpDomain(cmpDomain);
         assertEquals(cmpDomain, consentLibBuilder.cmpDomain);
     }
 
     @Test
-    public void testSetTargetingParamString() throws Exception {
+    public void setTargetingParamString() throws Exception {
         String key = "key";
         String stringValue = "stringValue";
-        EncodedParam targetingParamsString;
         consentLibBuilder.setTargetingParam(key, stringValue);
 
         Field localMember = getDeclaredFieldAccess("targetingParams");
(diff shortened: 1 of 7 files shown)


Suryakantj:
changes from 3rd commit are for gdpr status check, previous 2 commits are related to unit test cases writing.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @Before
+    public void initConsentLibBuilder() {
+        consentLibBuilder = new ConsentLibBuilder(808, "cybage.com", mock(Activity.class));
+        callback = new ConsentLib.Callback() {
+            @Override
+            public void run(ConsentLib c) {
+
+            }
+        };
+    }
```
andresilveirah: Really cool, using the `@Before` annotation make the test cases much more readable!
Just 2 small notes:
1. I'd change the account id from `808` to `123` and site name from `cybage.com` to `example.com` just to imply that this code being tested doesn't depend on a specific account id or site name.
2. I think we can refactor the `callback` assignment to pass a lambda instead. Something like:
```java
callback = (ConsentLib c) -> {};
```
Just so we have less code while keeping it still readable :)

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Test(expected = ConsentLibException.class)
+    public void testBuild() throws ConsentLibException {
+        ConsentLib consentLib = mock(ConsentLib.class);
+        assertEquals(consentLib, consentLibBuilder.build());
+    }
+
+    @Test
+    public void testSetMessageTimeout() {
+        consentLibBuilder.setMessageTimeOut(20000);
```
andresilveirah: I've checked a few other [testing examples](https://github.com/googlesamples/android-testing/blob/master/runner/AndroidJunitRunnerSample/app/src/androidTest/java/com/example/android/testing/androidjunitrunnersample/CalculatorTest.java) and I believe we have to rename the method names to something related to what's being tested in that method.  For example:
Instead of 
```java
 @Test
    public void testSetMessageTimeout() 
```
We should name it 
```java
 @Test
    public void setMessageTimeout() {
```

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+        assertEquals(jsonObject.toString() , localMember.get(consentLibBuilder).toString());
+
+        targetingParamsString = new EncodedParam("targetingParams", jsonObject.toString());
+        Method method = ConsentLibBuilder.class.getDeclaredMethod("setTargetingParamsString");
+        method.setAccessible(true);
+        method.invoke(consentLibBuilder);
+        assertEquals(targetingParamsString.toString(), consentLibBuilder.targetingParamsString.toString());
+
+    }
```
andresilveirah: Let's keep one assertion per test case. So here I'd break this test case into two:

```java
@Test
public void setTargetingParamString() throws Exception {
  // test that the targeting param is indeed set
}

@Test
public void targetingParamStringIsEncoded() throws Exception {
  // test the value gets transformed into EncodedParam
}
```

We might still have some duplication this way, but then we have test cases which are very focused. If they break we'll know right away what's the assertion that broke.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/EncodedParamTest.java
```diff
@@ -0,0 +1,45 @@
...
+        method.setAccessible(true);
+        assertEquals(URLEncoder.encode(attrValue,encodeValue),method.invoke(encodedParam,attrName,attrValue));
+
+        Field member = EncodedParam.class.getDeclaredField("value");
+        member.setAccessible(true);
+        EncodedParam encodedParam1 = new EncodedParam(attrName,attrValue);
+
+        assertNotNull(member.get(encodedParam1).toString());
+        assertEquals(member.get(encodedParam1).toString(),encodedParam1.toString());
+    }
```
andresilveirah: Maybe we should reuse the `getDeclaredFieldAccess(String fieldName)` method here as well?
Alternatively, we can also consider adding public getters and setters to the class, so you don't have to test private fields/methods. What you think?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Before
+    public void setUp() throws ConsentLibException {
+        sourcePointClientBuilder = new SourcePointClientBuilder(808, "siteName", true);//mock(SourcePointClientBuilder.class);
+    }
+
+    private Field getDeclareFieldAccess(String fieldName) throws Exception{
+        Field member = SourcePointClientBuilder.class.getDeclaredField(fieldName);
+        member.setAccessible(true);
+        return member;
+    }
```
andresilveirah: What if we added the `getDeclaredFieldAccess(String fieldName)` to a helper class/module, so we don't need to re-implement it in every test class? This way we just have to implement it once and import it whenever needed.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
```diff
@@ -0,0 +1,205 @@
...
+                    listener.onSuccess(200, headers, jsonObject);
+                } else {
+                    Header[] headers = new Header[]{};
+                    listener.onFailure(404, headers, "Error", new ConsentLibException("Error"));
+                }
+
+                return null;
+            }
+        }).when(http).get(anyString(), any(ResponseHandlerInterface.class));
+    }
```
andresilveirah: It seems that the `doAPICallWithAnswer `, `doGetGDPRStatus `, `doGetCustomConsents` have a lot in common. The only difference is in the response. Maybe we can refactor that into a method and receive the response as a parameter?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
```diff
@@ -0,0 +1,205 @@
...
+        reset(onLoadComplete);
+
+        doAPICallWithAnswer(false);
+
+        sourcePointClient.getSiteID(onLoadComplete);
+
+        verify(onLoadComplete, never()).onSuccess(any());
+
+        verify(onLoadComplete, times(1)).onFailure(any(ConsentLibException.class));
+    }
```
andresilveirah: I'd break each of these test cases into two. One for success and one for failure. So, instead of 
```java
@Test
    public void getSiteID() 
```
We'd have;
```java
@Test
    public void getSiteIDOnSuccess() {}

@Test
    public void getSiteIDOnFailure() {}
```

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        doGetCustomConsents(false);
+
+        sourcePointClient.getSiteID(onLoadComplete);
+
+        verify(onLoadComplete, never()).onSuccess(any());
+
+        verify(onLoadComplete, times(1)).onFailure(any(ConsentLibException.class));
+    }
+}
```
andresilveirah: I noticed the tests are only performed on mock objects. That's fine when we want to test other functionalities, like the `onSuccess` and `onFailure` callbacks are called and so on. 
But we need to test the real endpoint as well. That is, for example, when calling `getSiteId` for a specific account id and site name, we'll get an expected response. Maybe that's more like a "integration" test rather than unit testing. Does that make sense?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
Suryakantj: good suggestion , I would refactor code as suggested.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
Suryakantj: I would rename the methods by removing "test" prefix.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/EncodedParamTest.java
Suryakantj: I think we should not change access specifier just for the test case purpose and we should maintain test code each class in single file.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientBuilderTest.java
Suryakantj: We should maintain test code of each class in single file.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
Suryakantj: good suggestion , will refactor all code in single method.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
Suryakantj: will refactor to one assertion per method

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/ConsentLibBuilderTest.java
Suryakantj: refactored the code to one assertion per method

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
```diff
@@ -33,33 +33,31 @@
...
-    private void doAPICallWithAnswer(final boolean onSuccess) {
+    private void doAPICallWithAnswer(final boolean onSuccess , String response) {
         doAnswer(new Answer() {
             @Override
             public Object answer(InvocationOnMock invocation) {
-                System.out.println("TTTT : CALLED");
                 SourcePointClient.ResponseHandler listener = (SourcePointClient.ResponseHandler) invocation.getArguments()[1];
 
                 if (onSuccess) {
                     String stringToParse = "{\"site_id\":\"http://google.com\"}";
```
andresilveirah: Since now we're using the `response` parameter, do we still need this line?
```java
String stringToParse = "{\"site_id\":\"http://google.com\"}";
```

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
```diff
@@ -77,129 +75,80 @@ public Object answer(InvocationOnMock invocation) {
...
+    public void getSiteIDFailure(){
+        String response = "{\"site_id\":\"http://google.com\"}";
+        doAPICallWithAnswer(false,response);
 
         sourcePointClient.getSiteID(onLoadComplete);
 
         verify(onLoadComplete, never()).onSuccess(any());
-
         verify(onLoadComplete, times(1)).onFailure(any(ConsentLibException.class));
+        reset(onLoadComplete);
```
andresilveirah: What if we move the line:
```java
reset(onLoadComplete);
```
to a `@After`?
```java
@After
public void resetOnLoadCompleteMock() {
  reset(onLoadComplete);
}
```
Would that work?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
Suryakantj: I will remove this string, its not required anymore

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
Suryakantj: yes, that would also work.

Suryakantj: Refactored the code

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SourcePointClientTest.java
Suryakantj: removed the unused string variable

PR #68, opened 2019-12-11
changes made to resolve memeory leak issue

Suryakantj:
1. released activity reference in onDestroy method
2. created ConsentLibInstance class to return single instance of consentLib.
3. implemented onSaveInstanceState and onRetanInstanceState methods of activity to reuse consenLib object.
These changes restrict multiple object creation but application crashes on screen rotation.


Suryakantj:
these changes lead to an application crash , so closing this PR.

PR #69, opened 2019-12-16
Native clientJs implementation

leodoin:
## Remove JavaScript Client and handle actions natively
This is a major re-write of our SDK. We have now removed as much logic as possible from the WebView and put it in the hands of our Native code. The WebView is used now purely as a view layer, communicating with the SDK via `WKScriptMessageHandler` methods. 

### The Advantages
* Loading a message is now much faster and deterministic.
* The code is more maintainable and debuggable.
On top of that, this major refactoring is also a pre-requirement for our next step towards fully native consent messaging. 

### Disadvantages
0

### Status
- [x] Get message from Wrapper API (Mocked)
- [x] Inject Message events listener into WebView
- [x] Inject PM events listener into WebView
- [ ] Handle iAB variables on POST response
- [ ] Un-mock [POST action to Wrapper API](https://gist.github.com/andresilveirah/735e42319668f726185fae75d0679575#post-action)
- [ ] Un-mock message response from [GET Wrapper API](https://gist.github.com/andresilveirah/735e42319668f726185fae75d0679575#get-messagemandatoryparamsoptionalparams)
- [ ] Double check error handling and Internet Connectivity
- [ ] Fix Layout


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/ConsentLib.java
```diff
@@ -208,22 +220,93 @@ public void onConsentReady(String euConsent, String consentUUID) {
...
                 switch (choiceType) {
-                    case 12:
+                    case ActionTypes.SHOW_PM:
                         ConsentLib.this.choiceType = MESSAGE_OPTIONS.SHOW_PRIVACY_MANAGER;
+                        onShowPm();
+                        break;
+                    case ActionTypes.MSG_ACCEPT:
+                        onMsgAccepted();
+                        break;
+                    case ActionTypes.MSG_CLOSE:
```
andresilveirah: Is `MSG_CLOSE` the dismiss action? If so, wouldn't it be better to call it `MSG_DISMISS`?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/ConsentLib.java
```diff
@@ -238,14 +321,38 @@ public void onMessageChoiceSelect(int choiceId, int choiceType) {
...
+    private void renderMessage() throws ConsentLibException.NoInternetConnectionException {
+        if(webView == null) { webView = buildWebView(); }
+        sourcePoint.getMessageParams(new OnLoadComplete() {
+            @Override
+            public void onSuccess(Object result) {
+                try{
+                    String  msgUrl = ((JSONObject) result).getString("url");
+                    webView.loadUrlWithException(msgUrl);
+                    mCountDownTimer = getTimer(defaultMessageTimeOut);
+                    mCountDownTimer.start();
```
andresilveirah: I think this timer needs to be outside `getMessageParams`. We start a timer when the SDK starts running and we stop it either when the message shows up or when there's no message to show.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/ConsentWebView.java
```diff
@@ -250,7 +316,7 @@ public boolean onCreateWindow(WebView view, boolean dialog, boolean userGesture,
             }
             return false;
         });
-        addJavascriptInterface(new MessageInterface(), "JSReceiver");
```
andresilveirah: I first thought about renaming `JSReceiver` to `SDK` as well. But I think having a different name to reference the javascript code that our SDK injects in the WebView is useful and I decided to keep `JSReceiver`. What you think?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/ConsentLib.java
leodoin: Agreed, I din't do with the timer yet, just copied past this code by now.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/ConsentWebView.java
leodoin: lol I also think so, was just using SDK to try to be concise with the IOs naming as well.

PR #75, opened 2020-01-29
fix iabConsntString saved with consentUUID value

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/NativeView.java
```diff
@@ -0,0 +1,11 @@
...
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.View;
+
+public class NativeView extends View {
+    public NativeView(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+}
```
andresilveirah: @leodoin I think we should remove this file from this PR.

cmplibrary/src/main/res/values/attr.xml
```diff
@@ -0,0 +1,9 @@
+<resources>
+    <declare-styleable name="NativeView">
+        <attr name="showText" format="boolean" />
+        <attr name="labelPosition" format="enum">
+            <enum name="left" value="0"/>
+            <enum name="right" value="1"/>
+        </attr>
+    </declare-styleable>
+</resources>
```
andresilveirah: same for this one

PR #89, opened 2020-02-18
Tcfv2 support

leodoin:


app/src/main/java/com/sourcepoint/test_project/MainActivity.java
```diff
@@ -36,6 +36,7 @@ private void removeView(View view) {
     private GDPRConsentLib buildGDPRConsentLib() {
         return GDPRConsentLib.newBuilder(22, "a-demo-property", 7055,"5c0e81b7d74b3c30c6852301",this)
                 .setStagingCampaign(false)
+                .setAuthId("17-02-2020-0001")
```
andresilveirah: I believe we should remove the `authId`

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java
```diff
@@ -48,36 +55,34 @@
...
-        editor.commit();
-    }
-
-    public void setIabConsentParsedPurposeConsents(String consentParsedPurposeConsents){
-        editor.putString(IAB_CONSENT_PARSED_PURPOSE_CONSENTS, consentParsedPurposeConsents);
+    public void setTCData(HashMap<String, String> data){
+        for(String s : data.keySet()){
+            editor.putString(s, data.get(s));
+        }
+        editor.putString(TC_KEYS_KEY, dataSetToStr(data.keySet()));
```
andresilveirah: Why do we need the `TC_KEYS_KEY`?
So we can remove all IAB data later on?

leodoin: Yes, and also in order to get the TCData back from the local storage.

app/src/main/java/com/sourcepoint/test_project/MainActivity.java
leodoin: indeed

PR #99, opened 2020-03-06
Removed warnings from metaapp

Suryakantj:
Removed warnings from metaapp


Suryakantj:
@leodoin Build is successful , test cases executed successfully.

PR #104, opened 2020-03-25
Native msg font size as int

leodoin:
* font size fixed (now user can see different font sizes on the screen)
* crash on msg dismiss at first launch fixed
* can receive font size as int from msgJSON

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
@@ -44,7 +44,7 @@ public class GDPRConsentLib {
 
     public ConsentLibException error = null;
 
-    public GDPRUserConsent userConsent;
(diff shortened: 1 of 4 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -468,7 +468,7 @@ private void onErrorTask(ConsentLibException e){
     private void storeData(){
         storeClient.setConsentUuid(consentUUID);
         storeClient.setMetaData(metaData);
-        storeClient.setTCData(userConsent.TCData);
+        if(userConsent != null) storeClient.setTCData(userConsent.TCData);
```
andresilveirah: We could initialise `userConsent` with empty values and get rid of this null check. What you think?

leodoin: done

PR #114, opened 2020-04-02
Store IAB data on get message

andresilveirah:
Fixes https://sourcepoint.atlassian.net/browse/SP-3253

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
@@ -335,13 +335,14 @@ public class GDPRConsentLib {
                     JSONObject jsonResult = (JSONObject) result;
                     consentUUID = jsonResult.getString("uuid");
                     metaData = jsonResult.getString("meta");
+                    userConsent = new GDPRUserConsent(jsonResult.getJSONObject("userConsent"));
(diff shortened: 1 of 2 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -335,13 +335,14 @@ public void onSuccess(Object result) {
...
                     metaData = jsonResult.getString("meta");
+                    userConsent = new GDPRUserConsent(jsonResult.getJSONObject("userConsent"));
+                    storeData();
                     if(jsonResult.has("msgJSON") && !jsonResult.isNull("msgJSON")) {
                         setNativeMessageView(jsonResult.getJSONObject("msgJSON"));
                         showView(nativeView);
                     } else if(jsonResult.has("url") && !jsonResult.isNull("url")){
                         loadConsentUI(jsonResult.getString("url"));
                     } else {
-                        userConsent = new GDPRUserConsent(jsonResult.getJSONObject("userConsent"));
```
andresilveirah, line 344: @leodoin `userConsent` was only being assigned if there was not `msgJSON` or `url`. So, only in the second launch that'd be the case.

leodoin: got it

PR #128, opened 2020-04-19
Example app package rename

leodoin:
* app name on build.gradle was different from android.manifest so google play console was throwing and error when trying to release it
* app version code and app version name updated
*. minSdkVersion set to 21 (not sure if we wanna keep it as 16 for testing on google play console)


app/build.gradle
```diff
@@ -3,11 +3,11 @@ apply plugin: 'com.android.application'
 android {
     compileSdkVersion 28
     defaultConfig {
-        applicationId "com.example.dmitrirabinowitz.test_project"
-        minSdkVersion 16
+        applicationId "com.sourcepoint.example_app"
+        minSdkVersion 21
```
andresilveirah, line 7: Not sure we can bump from 16 to 21. Is it required by the PlayStore?

PR #132, opened 2020-04-24
SP-3438 Updated network calls library TCFv1

Suryakantj:
Implemented okhttp network call library which supports from android  API 9.


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -351,7 +351,13 @@ public void onSuccess(Object result) {
                         setNativeMessageView(jsonResult.getJSONObject("msgJSON"));
                         showView(nativeView);
                     } else if(jsonResult.has("url") && !jsonResult.isNull("url")){
-                        loadConsentUI(jsonResult.getString("url"));
+                        String url = jsonResult.getString("url");
```
leodoin, line 354: This change is not related to this ticket, is it?

Suryakantj: Yes it is related.

PR #134, opened 2020-04-25
SP-3071 more unit tests tcfv2

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ActionTypes.java
```diff
@@ -1,11 +1,26 @@
...
 
-public class ActionTypes {
-
-    public static final int SHOW_OPTIONS = 12;
-    public static final int REJECT_ALL = 13;
-    public static final int ACCEPT_ALL = 11;
-    public static final int MSG_CANCEL = 15;
-    public static final int SAVE_AND_EXIT = 1;
-    public static final int PM_DISMISS = 2;
+public enum ActionTypes {
```
andresilveirah, line 3: Should the name be in the singular? `ActionType`?

leodoin: Yes, its should, but it would be a breaking change as this this a public.

PR #135, opened 2020-04-30
SP-3436 dismiss action on back button press within message - TCFv2

Suryakantj:
implemented on back button press dismiss action

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
@@ -16,4 +16,9 @@ public class ConsentAction {
         this.requestFromPm = requestFromPm;
         this.pmSaveAndExitVariables = pmSaveAndExitVariables;
     }
+
+    public static ConsentAction getEmptyDismissAction(){
+        return new ConsentAction(15, "",false,new JSONObject());
+    }
+
 }
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
@@ -18,11 +18,8 @@ public class ConsentAction {
     }
 
     public static ConsentAction getEmptyDismissAction(boolean isPmOn){
-        if (isPmOn){
-            return new ConsentAction(2, "", true, new JSONObject());
-        }else {
-            return new ConsentAction(15, "", false, new JSONObject());
-        }
+        return isPmOn ? new ConsentAction(2, "", true, new JSONObject())
+                : new ConsentAction(15, "", false, new JSONObject());
     }
 
 }
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
@@ -17,8 +17,12 @@ public class ConsentAction {
         this.pmSaveAndExitVariables = pmSaveAndExitVariables;
     }
 
-    public static ConsentAction getEmptyDismissAction(){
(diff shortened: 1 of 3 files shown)


Suryakantj:
> Unfortunately this does not cover the case when the user presses the go-back button from the pm opened directly from the app.

Included pm dismiss action

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
```diff
@@ -16,4 +16,13 @@
         this.requestFromPm = requestFromPm;
         this.pmSaveAndExitVariables = pmSaveAndExitVariables;
     }
+
+    public static ConsentAction getEmptyDismissAction(boolean isPmOn){
```
leodoin, line 20: I think it's cleaner with a ternary comparison for this case:
  ```JAVA
return isPmOn ? new ConsentAction(2, "", true, new JSONObject()) : new ConsentAction(15, "", false, new JSONObject());
```
It's just a matter of style 😅 What do you think?

File: cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentWebView.java [hunk 2]
```diff
@@ -97,6 +97,7 @@ private void setup() {
         }
         getSettings().setJavaScriptEnabled(true);
         this.setBackgroundColor(Color.TRANSPARENT);
+        this.requestFocus();
```
leodoin, line 100: why did we need this here?

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -194,6 +194,16 @@ public void onError(ConsentLibException error) {
             public void onAction(ConsentAction action) {
                 GDPRConsentLib.this.onAction(action);
             }
+
+            @Override
+            public void onBackPressAction() {
+                if (webView.canGoBack()){
```
leodoin, line 200: here you should change `webview` by `this`

leodoin: you don't even need to call this.goBack() here. Just call `GDPRConsentLib.this.onAction(ConsentAction.getEmptyDismissAction(isPmOn));`
and let the magic happen :)

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentWebView.java
Suryakantj: We need this because onBackPressed event will not be listen by web-view until user interacts with it or web-view have focus.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
Suryakantj: will  do that.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
Suryakantj: Yeah , Good suggestion

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
```diff
@@ -16,4 +16,10 @@
...
+        return isPmOn ? new ConsentAction(2, "", true, new JSONObject())
```
leodoin, line 21: here use can use the actionTypes enum instead of the hard code.

Suryakantj: Good suggestion

PR #141, opened 2020-05-04
SP-3436 dismiss action on back press TCFv1

Suryakantj:
implemented onBack pressed dismiss action tcfv1


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -201,6 +201,15 @@ public void onSavePM(GDPRUserConsent u) {
             public void onAction(int choiceType, Integer choiceId) {
                 GDPRConsentLib.this.onAction(choiceType, choiceId);
             }
+
+            @Override
+            public void onBackPressAction() {
+                if (isPmOn){
+                    GDPRConsentLib.this.onAction(2,null);
```
leodoin, line 208: Here you should rather use ActionTypesClass instead of hardcoding the action types.

Suryakantj: Good suggestion

PR #142, opened 2020-05-04
Sp 3438 network calls library change tcfv1

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -344,14 +329,15 @@ private void renderMsgAndSaveConsent() throws ConsentLibException {
             @Override
             public void onSuccess(Object result) {
                 try{
-                    JSONObject jsonResult = (JSONObject) result;
+                    JSONObject jsonResult = new JSONObject((String) result);
```
andresilveirah, line 332: just curiosity, what's the difference between `(JSONObject) result` vs `new JSONObject((String) result)`. Were we getting casting exception?

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -344,14 +329,15 @@ private void renderMsgAndSaveConsent() throws ConsentLibException {
...
                     consentUUID = jsonResult.getString("uuid");
                     metaData = jsonResult.getString("meta");
                     if(jsonResult.has("msgJSON") && !jsonResult.isNull("msgJSON")) {
                         setNativeMessageView(jsonResult.getJSONObject("msgJSON"));
                         showView(nativeView);
                     } else if(jsonResult.has("url") && !jsonResult.isNull("url")){
-                        loadConsentUI(jsonResult.getString("url"));
+                        String url = jsonResult.getString("url");
+                        loadConsentUI(url);
```
andresilveirah, line 340: did you perhaps add the intermediate `url` variable for debugging and forgot inline back inside `loadConsentUI`?

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
leodoin: yes, were were getting casting exception. The JSONObject constructor explicitly parse the the string to a JSONObject while the dynamic cast does not, It simply try to match the casted class as a parent and in this case it breaks as those classes are not living under the same roof 😅

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
leodoin: I was actually testing your attention there =P
Gonna change it 😅

PR #144, opened 2020-05-06
SP-3534 refactor store client

andresilveirah:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java
```diff
@@ -110,25 +83,14 @@ public void clearAllData(){
...
-    private void clearTCData() {
-        for(String s : dataSetFromStore()){
-            editor.remove(s);
-        }
-        editor.remove(TC_KEYS_KEY);
-    }
-
-    private String[] dataSetFromStore(){
-        return pref.getString(TC_KEYS_KEY, "").split(TC_KEYS_DELIMITER);
+        for(String key : pref.getAll().keySet()) if (key.startsWith(IABTCF_KEY_PREFIX))
```
leodoin, line 92: It's really just a matter of taste, but don't you think it looks better this way:
```Java
for(String key : pref.getAll().keySet()) 
  if (key.startsWith(IABTCF_KEY_PREFIX)) editor.remove(key);
```

PR #145, opened 2020-05-06
add first unit test for async behaviour -> onShowOptions()

leodoin:


cmplibrary/src/test/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLibTest.java
```diff
@@ -117,8 +123,10 @@ public void onAction(){
     }
 
     @Test
-    public void onShowOptions(){
+    public void onShowOptions() {
         lib.onShowOptions();
+        verify(lib.activity).runOnUiThread(lambdaCaptor.capture());
+        lambdaCaptor.getValue().run();
```
andresilveirah, line 129: Jees, if you didn't explain me what this code is doing I'd never guess... what an ugly API 🤢 
(just to be clear, your code is not ugly, it's Mockito + Junit + Java in general.

PR #146, opened 2020-05-07
Sp 3439 network calls library change tcfv2 rebased

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/SourcePointClient.java
```diff
@@ -2,112 +2,75 @@
...
+        Request request = new Request.Builder().url(url).post(body)
+                .header("Accept", "application/json")
+                .header("Content-Type", "application/json")
+                .build();
+
+        httpClient.newCall(request).enqueue(new Callback() {
+            @Override
+            public void onFailure(Call call, IOException e) {
+                Log.d(LOG_TAG, "Failed to load resource " + url + " due to " +   "url load failure :  " + e.getMessage());
+                onLoadComplete.onFailure(new ConsentLibException(e, "Fail to send consent to: " + url));
```
andresilveirah, line 59:
```java
onLoadComplete.onFailure(new ConsentLibException(e, "Fail to send consent to: " + url));
```
Maybe the reason `Fail to send consent to` is not accurate here as we're not sending consent but getting the message?

andresilveirah: This might be a _copy n paste_ mistake as the code seems similar from below.

leodoin: yeap, gonna change it 👍

PR #156, opened 2020-05-18
Sp 3616 check intenet connection

Suryakantj:
Added network connection check before making http call and loading url

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
@@ -256,9 +256,13 @@ public class GDPRConsentLib {
         nativeView.setAttributes(new NativeMessageAttrs(msgJson));
     }
 
-    public void onMsgAccepted(Integer choiceId) throws ConsentLibException{
+    public void onMsgAccepted(Integer choiceId) {
         closeAllViews();
-        sendConsent(ActionTypes.MSG_ACCEPT, choiceId);
+        try {
+            sendConsent(ActionTypes.MSG_ACCEPT, choiceId);
+        } catch (ConsentLibException e) {
+            onErrorTask(e);
+        }
     }
 
     protected void onPmDismiss() {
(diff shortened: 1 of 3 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -249,7 +256,7 @@ private void setNativeMessageView(JSONObject msgJson) throws ConsentLibException
         nativeView.setAttributes(new NativeMessageAttrs(msgJson));
     }
 
-    public void onMsgAccepted(Integer choiceId) {
+    public void onMsgAccepted(Integer choiceId) throws ConsentLibException{
```
leodoin, line 259: We should not change the signature of any public methods, here we can catch the exception instead

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -411,36 +430,40 @@ private JSONObject paramsToSendConsent(int actionType, Integer choiceId) throws
...
-                        consentUUID = jsonResult.getString("uuid");
-                        metaData = jsonResult.getString("meta");
-                        Log.i("GDPR_UUID", "From sendConsentReponse: " + consentUUID);
-                        consentFinished();
-                    } catch (Exception e) {
-                        Log.d(TAG, "Sorry, something went wrong");
-                        e.printStackTrace();
-                        onErrorTask(new ConsentLibException(e, "Error trying to parse response from sendConsents."));
+    protected void sendConsent(int actionType, Integer choiceId) throws  ConsentLibException {
+        if (!hasLostInternetConnection()) {
```
leodoin, line 434: we throw the exception directly instead of using a else clause

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -278,26 +285,30 @@ public void onMsgCancel(Integer choiceId) {
...
     }
 
-    private void loadConsentUI(String url) {
-        mCountDownTimer.start();
-        runOnLiveActivityUIThread(() -> {
-            if (webView == null) webView = buildWebView();
-            webView.loadConsentUIFromUrl(url);
-        });
+    private void loadConsentUI(String url) throws  ConsentLibException {
+        if (!hasLostInternetConnection()) {
```
leodoin, line 303: here we can also throw instead of checking having a else clause

Suryakantj: will do this and follow same guidelines for future tasks

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
Suryakantj: yes will do this and follow same guidelines in future task.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
Suryakantj: will do this.

PR #160, opened 2020-05-25
SP-3686 example crash due to empty userconsent object

Suryakantj:
fixed example crash due to empty userconsent object

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRUserConsent.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRUserConsent.java
@@ -10,13 +10,15 @@ public class GDPRUserConsent {
 
     public ArrayList<String> acceptedVendors;
     public ArrayList<String> acceptedCategories;
-    public JSONObject jsonConsents;
+    public JSONObject jsonConsents = new JSONObject();
     public String consentString;
 
-    private static JSONObject emptyConsents = null;
-
     public GDPRUserConsent() throws JSONException {
-        this(getEmptyUserConsents());
+        this.acceptedVendors = new ArrayList<>();
+        this.acceptedCategories = new ArrayList<>();
+        this.consentString = "";
+        jsonConsents.put("acceptedVendors",new JSONArray());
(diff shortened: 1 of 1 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRUserConsent.java
```diff
@@ -41,4 +41,15 @@ public GDPRUserConsent(JSONObject jConsent) throws JSONException, ConsentLibExce
     public JSONObject getJsonConsents() {
         return jsonConsents;
     }
+
+    private static JSONObject getEmptyUserConsents() throws JSONException{
```
leodoin, line 45: Do we really need a static method here? It should be avoided whenever possible and I think this is the case here.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRUserConsent.java
```diff
@@ -13,13 +13,13 @@
     public JSONObject jsonConsents;
     public String consentString;
 
-    public GDPRUserConsent(){
-        acceptedVendors = new ArrayList<>();
-        acceptedCategories = new ArrayList<>();
-        consentString = "";
+    private static JSONObject emptyConsents = null;
+
+    public GDPRUserConsent() throws JSONException {
```
leodoin, line 18: We should not use JSONObject here, we can just set the jsonConsent field directly, no exception needed!

Suryakantj: we need exception here , as we are putting empty values in jsonConsent object , otherwise we will get exception in application.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRUserConsent.java
Suryakantj: yes we can remove static method from here and initialization logic in constructor itself

PR #167, opened 2020-06-02
SP-3627 add cached user consent

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/CustomJsonParser.java
```diff
@@ -1,10 +1,13 @@
 package com.sourcepoint.gdpr_cmplibrary;
 
+import android.app.MediaRouteActionProvider;
```
andresilveirah, line 3: maybe this dependency was added by mistake?

leodoin: @andresilveirah thanks for the note. Dependence cleared 🕺

PR #174, opened 2020-06-10
SP 3197 implement lifecycle methods from iOS

Suryakantj:
Added  pmWillShow , pmDidDisappear, messageWillShow and messageDidDisappear lifecycle methods.

### DIFF
--- a/app/src/main/java/com/sourcepoint/example_app/MainActivity.java
+++ b/app/src/main/java/com/sourcepoint/example_app/MainActivity.java
@@ -80,6 +80,10 @@ public class MainActivity extends AppCompatActivity {
                     Log.e(TAG, "Something went wrong: ", error);
                     Log.i(TAG, "ConsentLibErrorMessage: " + error.consentLibErrorMessage);
                 })
+                .setPMWillShow(()-> Log.e(TAG, "PMWillShow"))
+                .setMessageWillShow(()-> Log.e(TAG, "MessageWillShow"))
+                .setPMDidDisappear(()-> Log.e(TAG, "PMDidDisappear"))
+                .setMessageDidDisappear(()-> Log.e(TAG, "MessageDidDisappear"))
                 .build();
     }
 
--- a/app/src/main/java/com/sourcepoint/example_app/MainActivity.java
+++ b/app/src/main/java/com/sourcepoint/example_app/MainActivity.java
@@ -80,10 +80,10 @@ public class MainActivity extends AppCompatActivity {
                     Log.e(TAG, "Something went wrong: ", error);
                     Log.i(TAG, "ConsentLibErrorMessage: " + error.consentLibErrorMessage);
                 })
-                .setPMWillShow(()-> Log.e(TAG, "PMWillShow"))
-                .setMessageWillShow(()-> Log.e(TAG, "MessageWillShow"))
-                .setPMDidDisappear(()-> Log.e(TAG, "PMDidDisappear"))
-                .setMessageDidDisappear(()-> Log.e(TAG, "MessageDidDisappear"))
+                .setPMReady(()-> Log.e(TAG, "PM Ready"))
+                .setMessageReady(()-> Log.e(TAG, "Message Ready"))
+                .setPMFinished(()-> Log.e(TAG, "PM Finished"))
+                .setMessageFinished(()-> Log.e(TAG, "Message Finished"))
                 .setOnAction(actionType  -> Log.e(TAG , "ActionType : "+actionType.toString()))
                 .build();
     }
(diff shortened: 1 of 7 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -50,6 +50,11 @@
     final OnConsentUIFinishedCallback onConsentUIFinished;
     final OnConsentReadyCallback onConsentReady;
     final OnErrorCallback onError;
+    final pmWillShowCallback pmWillShow;
```
leodoin, line 53: should rename callbacks to android standards here as well

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -354,32 +390,58 @@ public void onFailure(ConsentLibException exception) {
         });
     }
 
-    void showView(View view) {
+    void showView(View view, boolean isFromPM) {
         mCountDownTimer.cancel();
         if (!hasParent(view)) {
             runOnLiveActivityUIThread(() -> GDPRConsentLib.this.onConsentUIReady.run(view));
         }
+        if (isFromPM) runPMWillShow();
```
leodoin, line 398: could try make a ternary inline check here:
```Java
isFromPM ? runPMWillShow() : runMessageWillShow();
```

leodoin: Also, I don't think this code should be part of this method. It makes forces you to pass this parameter 'isFromPM' all the way down. Instead you could could just call either runPMWillShow() or runMessageWillShow() when needed. Does it make sense to you?

cmplibrary/src/main/res/raw/js_receiver.js
```diff
@@ -17,7 +17,8 @@ function consentData(event) {
 };
 
 function isFromPM(event) {
-    return !!event.data.payload;
+    event.data.settings = event.data.settings || {};
```
leodoin, line 20: here you don't have to change the original obj. Instead you can try a inline check like this:
```Javascript
return event.data.fromPM || event.data.settings && event.data.settings.vendorList;
```

app/src/main/java/com/sourcepoint/example_app/MainActivity.java
```diff
@@ -80,6 +80,11 @@ private GDPRConsentLib buildGDPRConsentLib() {
                     Log.e(TAG, "Something went wrong: ", error);
                     Log.i(TAG, "ConsentLibErrorMessage: " + error.consentLibErrorMessage);
                 })
+                .setPMWillShow(()-> Log.e(TAG, "PMWillShow"))
```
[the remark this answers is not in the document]
Suryakantj: I just copied from iOS to have same names , Now I have changed as per android convention

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
Suryakantj: We need this code here only as we need to have these callbacks after onConsentUIReady. as per description in ticket.

Suryakantj: > could try make a ternary inline check here:
> 
> ```java
> isFromPM ? runPMWillShow() : runMessageWillShow();
> ```

We can only perform inline ternary operation with assignment. 
Here we are not assigning value to any variable also these are void methods

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
Suryakantj: Renamed callback methods as per android convention

cmplibrary/src/main/res/raw/js_receiver.js
Suryakantj: Yes we can do this , also we need to have null check for vendorList otherwise this will return false always.

Suryakantj:
> Nice work! I have just requested a few style/refac changes, but the bulk of the work is pretty solid :)

Thank you!

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -354,32 +385,58 @@ public void onFailure(ConsentLibException exception) {
         });
     }
 
-    void showView(View view) {
+    void showView(View view, boolean isFromPM) {
```
leodoin, line 388: @Suryakantj , pls take this second argument out of this method, as i have comment on this PR already.

PR #175, opened 2020-06-11
Fixed MetaApp flow and increased default timeout to 30 sec.

Suryakantj:
fixed metaApp flow while creating new property and increased default timeout to 30 sec

### DIFF
--- a/metaapp/src/main/java/com/sourcepointmeta/app/ui/ConsentViewActivity.java
+++ b/metaapp/src/main/java/com/sourcepointmeta/app/ui/ConsentViewActivity.java
@@ -93,7 +93,7 @@ public class ConsentViewActivity extends BaseActivity<ConsentViewViewModel> {
     private GDPRConsentLib buildConsentLib(Property property, Activity activity) {
         ConsentLibBuilder consentLibBuilder = GDPRConsentLib.newBuilder(property.getAccountID(), property.getProperty(), property.getPropertyID(), property.getPmID(), activity)
                 .setStagingCampaign(property.isStaging())
-                //.setMessageTimeOut(15000)
+                .setMessageTimeOut(30000)
                 .setOnConsentUIReady(view -> {
                             getSupportActionBar().hide();
                             isPMLoaded = true;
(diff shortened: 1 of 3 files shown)


metaapp/src/main/java/com/sourcepointmeta/app/ui/ConsentViewActivity.java
```diff
@@ -93,7 +93,7 @@ private void removeWebView(View view) {
     private GDPRConsentLib buildConsentLib(Property property, Activity activity) {
         ConsentLibBuilder consentLibBuilder = GDPRConsentLib.newBuilder(property.getAccountID(), property.getProperty(), property.getPropertyID(), property.getPmID(), activity)
                 .setStagingCampaign(property.isStaging())
-                //.setMessageTimeOut(15000)
+                .setMessageTimeOut(30000)
```
leodoin, line 96: @Suryakantj I think it would be better to have a final static field for this default value. What do you think?

metaapp/src/main/java/com/sourcepointmeta/app/ui/NewPropertyActivity.java
```diff
@@ -94,12 +92,11 @@ private void removeWebView(View view) {
     private GDPRConsentLib buildConsentLib(Property property, Activity activity) {
         ConsentLibBuilder consentLibBuilder = GDPRConsentLib.newBuilder(property.getAccountID(), property.getProperty(), property.getPropertyID(), property.getPmID(), activity)
                 .setStagingCampaign(property.isStaging())
-                .setMessageTimeOut(15000)
+                .setMessageTimeOut(30000)
```
leodoin, line 95: preferably we should recover this default value from the same single static source

metaapp/src/main/java/com/sourcepointmeta/app/ui/ConsentViewActivity.java
Suryakantj: Good suggestion.

metaapp/src/main/java/com/sourcepointmeta/app/ui/NewPropertyActivity.java
Suryakantj: Done

PR #179, opened 2020-06-17
SP-3606 listen to error event on JSReceiver

leodoin:


cmplibrary/src/test/JS/test_config.js
```diff
@@ -0,0 +1,54 @@
+var relativePath = '../../main/res/raw/js_receiver.js'
+
+var eventMocks = {
```
andresilveirah, line 3: nice 💃

PR #181, opened 2020-06-19
[TCFv1] Added initial automation test suite

deshpandevrushali:


leodoin:
@deshpandevrushali , could you verify why is this not passing the tests ? If it takes to long we can simply close the PR since TCFV1 is not a priority anymore.

deshpandevrushali:
> @deshpandevrushali , could you verify why is this not passing the tests ? If it takes to long we can simply close the PR since TCFV1 is not a priority anymore.

@leodoin I think as we can close this PR since TCFV1 is not a priority anymore.

PR #182, opened 2020-06-19
SP-3790 Modified getTCData method and added tests for tcdata

Suryakantj:
Modified gtTCData method to get right primitiv value from sharedPref
Added unit tests to verify tcdata with premitives
Removed constants from StoreClientTest file.

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java
@@ -6,6 +6,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.HashMap;
+import java.util.Map;
 
 public class StoreClient {
 
@@ -111,13 +112,9 @@ public class StoreClient {
 
     public HashMap getTCData(){
         HashMap tcData = new HashMap();
+        Map map = pref.getAll();
         for(String key : pref.getAll().keySet()) if (key.startsWith(IABTCF_KEY_PREFIX))
-            try {
-                String value = pref.getString(key,null);
-                tcData.put(key, value);
-            }catch (ClassCastException e){
-                tcData.put(key , pref.getInt(key,-1));
-            }
+            tcData.put(key, map.get(key));
         return tcData;
     };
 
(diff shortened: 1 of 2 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java
```diff
@@ -112,7 +112,12 @@ public void clearInternalData(){
     public HashMap getTCData(){
         HashMap tcData = new HashMap();
         for(String key : pref.getAll().keySet()) if (key.startsWith(IABTCF_KEY_PREFIX))
-            tcData.put(key, pref.getString(key, null));
+            try {
```
leodoin, line 115: better to make a type check like we do [here](https://github.com/SourcePointUSA/android-cmp-app/blob/76e1c2dc6b402603e8fd7796c360083e5dc514f7/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java#L43).

cmplibrary/src/test/java/com/sourcepoint/gdpr_cmplibrary/StoreClientTest.java
```diff
@@ -101,25 +98,60 @@ public void getAuthId() {
...
+        storeClient.setTCData(tcData);
+
+        Map map = sharedPreferences.getAll();
+
+        assertEquals("foo", map.get("IABTCF_String"));
+        assertEquals(100 , map.get("IABTCF_Integer"));
+    }
+
+    @Test
+    public void getTCDataAndCheckPrimitives() {
```
leodoin, line 140: There is no need for this test since the one before does the job already

Suryakantj: @leodoin This is test method for getTcData , so we should have this test case.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/StoreClient.java
Suryakantj: @leodoin  I have updated this method to remove try catch block

PR #183, opened 2020-06-22
SP-4029 change activity ref for application context

leodoin:
ready 💯


andresilveirah:
bye bye `Activity` reference  👋

PR #199, opened 2020-07-08
SP-4222 Fixed onError callback called twice.

Suryakantj:
Fixed onError being called twice 
Also ensured once onError or OnConsentReady is called then sdk cycle is finished.


Suryakantj:
Closing this PR, as this issue been resolved,

PR #202, opened 2020-07-09
[SP-4122] Revert back button action override

leodoin:
and a little bit more...

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
@@ -226,10 +226,8 @@ public class GDPRConsentLib {
     }
 
     public void onAction(ConsentAction action) {
-        if (GDPRConsentLib.this.onAction != null)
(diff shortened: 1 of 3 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -226,10 +226,8 @@ public void onBackPressAction() {
     }
 
     public void onAction(ConsentAction action) {
-        if (GDPRConsentLib.this.onAction != null)
-            uiThreadHandler.postIfEnabled( () ->GDPRConsentLib.this.onAction.run(action.actionType));
-
         try {
+            uiThreadHandler.postIfEnabled( () ->GDPRConsentLib.this.onAction.run(action.actionType));
```
Suryakantj, line 230: Here we may get NPE if application implementing SDK has not initialised onAction callback

leodoin: the builder initialises it as a noop callback

PR #211, opened 2020-07-20
SP-4274 convert choice id tostring on js receiver

leodoin:


andresilveirah:
@leodoin is the commit history of this PR correct? I see some files I've added directly to `develop` in the list of changed files. For example, the GH workflow files.

PR #213, opened 2020-07-21
SP-4274 convert choice id to string on js receiver

leodoin:


cmplibrary/src/main/res/raw/js_receiver.js
```diff
@@ -24,7 +24,7 @@ function dataFromMessage(msgEvent) {
     return {
         name: msgEvent.data.name,
         actionType: msgEvent.data.actions.length ? msgEvent.data.actions[0].data.type : null,
-        choiceId: msgEvent.data.actions.length ? msgEvent.data.actions[0].data.choice_id : null,
```
andresilveirah, line 27: Should we add one unit test for the case in which the `choiceId` is `Number` or `String`? Specially now that we're jesting the shit out of our JSReceiver

PR #215, opened 2020-07-22
SP-4289 change user consent fields to interface types

leodoin:


andresilveirah:
@Suryakantj can you have a look at the failing spec for us?

Suryakantj:
> @Suryakantj can you have a look at the failing spec for us?

@andresilveirah I have updated couple of MetaApp files to accept collection from user consent object.
It's all good now.

PR #217, opened 2020-07-27
:100: test coverage for UIThreadHandler class

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/UIThreadHandler.java
```diff
@@ -21,7 +21,7 @@ public void enable(){
     }
 
     public boolean postIfEnabled(Runnable r){
-        return isEnabled ? super.post(r) : false;
+        return isEnabled ? post(r) : false;
```
Suryakantj, line 24: Do we need this return value?

leodoin: we don't use it on our  code but it is in line w/ the `android.Handler.post()` that we are extending

PR #220, opened 2020-07-28
SP-4304 Expose cached user consent

Suryakantj:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -583,6 +583,10 @@ void storeData() throws JSONException, ConsentLibException {
         storeClient.setUserConsents(userConsent);
     }
 
+    public GDPRUserConsent getCachedUserConsent() throws ConsentLibException {
```
leodoin, line 586: any method from consentLib should throw exception, best here would be to add a try/catch block and call onError instead

Suryakantj: @leodoin Calling onError will set uiThreadhandler isEnable flag to false  and impact will be no callback gets triggered when consentlib.run method is executed.

PR #225, opened 2020-07-30
SP-4304 static method for getting user consent from shared pref + StoreClient 100% coverage/refac

leodoin:


hameno:
This has been removed again, what is the current way to get the user consent?

PR #227, opened 2020-08-05
SP-4310 Automate meta app release using GitHub action

Suryakantj:
Updated Meta App package name 
Added release.yml file for auto release process of Meta App to play store


andresilveirah:
@Suryakantj I see we moved from `com.sourcepointmeta.app` to `com.sourcepointmeta.metaapp`. Could we instead, move to `com.sourcepoint.metaapp`?

Suryakantj:
> @Suryakantj I see we moved from `com.sourcepointmeta.app` to `com.sourcepointmeta.metaapp`. Could we instead, move to `com.sourcepoint.metaapp`?

@andresilveirah We already released  MetaApp on play store with this package name.

Suryakantj:
> I thing `app` should be renamed to `metaApp` (respecting camel case style) instead of `metaapp` or am I missing something?

package name is always lowercase letters

PR #247, opened 2020-08-21
[SP-4556] Implemented native UI tests for example app

deshpandevrushali:
Added 9 E2E native UI test for example app


leodoin (approved):
way to go 🎖️

PR #248, opened 2020-08-21
[SP-4603] Remove unused code and not needed null checks

Suryakantj:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -415,14 +408,12 @@ void showView(View view, boolean isFromPM) {
     }
 
     private void runPMReady(){
-        if (this.pmReady != null)
-            uiThreadHandler.postIfEnabled(GDPRConsentLib.this.pmReady::run);
+        uiThreadHandler.postIfEnabled(GDPRConsentLib.this.pmReady::run);
```
leodoin, line 411: @Suryakantj , have you tested this without setting the `onPmReady` callback ?

leodoin (approved):
To be on the safe side it's better to double check (right some tests) to ensure we are not throwing erros on the optional callback are not set by the user.

cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
Suryakantj: @leodoin I did test without setting those callbacks and it works as expected.

Suryakantj:
> To be on the safe side it's better to double check (right some tests) to ensure we are not throwing erros on the optional callback are not set by the user.

I have added unit tests for optional callback

leodoin (approved):
👍

PR #254, opened 2020-08-26
SP-4865 ensure consent lib exception does not get wrapped into another one

leodoin:


andresilveirah (approved):

andresilveirah:
We just need to remember to merge #252 before this one.

PR #260, opened 2020-08-26
SP-4114 implement pub data feature

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentLibBuilder.java
```diff
@@ -9,6 +9,8 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.concurrent.RunnableFuture;
```
andresilveirah, line 12: Was this added by mistake?

leodoin: yep, rookie mistake. got rid of it before merging though 👍

PR #263, opened 2020-08-28
[SP 4821] Handle null pubdata

Suryakantj:
We are getting error on defaultAction() due to null pubdata , Now we are having null check for pubdata.

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
@@ -36,7 +36,6 @@ public class ConsentAction {
     }
 
     public JSONObject getPubData(){
-        return new JSONObject(pubData);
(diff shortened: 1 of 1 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/ConsentAction.java
```diff
@@ -26,17 +26,11 @@
         this(actionType, choiceId, null, requestFromPm, pmSaveAndExitVariables);
     }
 
-    public static ConsentAction getEmptyDismissAction(boolean isPmOn) {
```
andresilveirah, line 29: Was this method deleted because it was unused?

Suryakantj: @andresilveirah Yes this method was unused.

Suryakantj:
Closed this one as we have covered this issue with another PR https://github.com/SourcePointUSA/android-cmp-app/pull/264

PR #269, opened 2020-09-14
[SP-4722] Get default PM tab from show pm action

Suryakantj:
Get default PM tab from show pm action from message and UI tests for example app & MetaApp

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
@@ -254,7 +254,7 @@ public class GDPRConsentLib {
             Log.d(TAG, "onAction:  " + action.actionType + " + actionType");
             switch (action.actionType) {
                 case SHOW_OPTIONS:
-                    onShowOptions(action.privacyManagerId);
+                    onShowOptions(action);
                     break;
                 case PM_DISMISS:
                     onPmDismiss(action.requestFromPm);
@@ -320,8 +320,8 @@ public class GDPRConsentLib {
         return isNative ? nativeView : webView;
     }
 
-    public void onShowOptions(String privacyManagerId) {
-        showPm(privacyManagerId);
+    public void onShowOptions(ConsentAction action) {
+        showPm(action.privacyManagerId, action.pmTab);
     }
 
     private void loadConsentUI(String url) {
@@ -354,14 +354,14 @@ public class GDPRConsentLib {
     }
 
     public void showPm() {
-        showPm(null);
+        showPm(null,null);
     }
 
-    public void showPm(String privacyManagerId) {
+    public void showPm(String privacyManagerId, String pmTab) {
         try {
             mCountDownTimer.start();
             isPmOn = true;
-            loadConsentUI(pmUrl(privacyManagerId));
+            loadConsentUI(pmUrl(privacyManagerId, pmTab));
         } catch (Exception e) {
             onErrorTask(new ConsentLibException(e, "Unexpected error on consentLib.showPm()"));
         }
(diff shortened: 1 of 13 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -354,14 +354,14 @@ public void run() {
     }
 
     public void showPm() {
-        showPm(null);
+        showPm(null,null);
```
leodoin, line 357: @Suryakantj , here would be better to overload the `showPm(String, String)` with a `showPm()` instead of calling it with null args

cmplibrary/src/main/res/raw/js_receiver.js
```diff
@@ -2,18 +2,25 @@ function isFromPM(payload) {
...
 
-function getPmIdFromURL(url) {
-    return url ? url.match(/[?&]message_id(=([^&#]*)|&|#|$)/)[2] : null;
+function getPmIdAndPmTabFromURL(url) {
+    let pmId = url ? url.match(/[?&]message_id(=([^&#]*)|&|#|$)/)[2] : null;
+    let pmTabShow = null;
+    if(url &&  url.match(/[?&]pmTab(=([^&#]*)|&|#|$)/) && url.match(/[?&]pmTab(=([^&#]*)|&|#|$)/)[2]){
+       pmTabShow = url.match(/[?&]pmTab(=([^&#]*)|&|#|$)/)[2];
+     }
+    return [pmId, pmTabShow];
```
leodoin, line 11: here you can return a object:
```JavaScript
{pmId,pmTabShow}
```
so later get the values by the keys instead :)

cmplibrary/src/main/res/raw/js_receiver.js
```diff
@@ -2,18 +2,25 @@ function isFromPM(payload) {
     return payload.fromPM || (payload.settings && payload.settings.vendorList != null);
 }
 
-function getPmIdFromURL(url) {
-    return url ? url.match(/[?&]message_id(=([^&#]*)|&|#|$)/)[2] : null;
+function getPmIdAndPmTabFromURL(url) {
+    let pmId = url ? url.match(/[?&]message_id(=([^&#]*)|&|#|$)/)[2] : null;
```
leodoin, line 6: avoid using ES6 features like `let` for ensuring cross webview/browser compatibility: use `var` instead

cmplibrary/src/main/res/raw/js_receiver.js
```diff
@@ -2,18 +2,25 @@ function isFromPM(payload) {
...
 function actionFromMessage(payload) {
     var actionPayload = payload.actions && payload.actions.length && payload.actions[0] && payload.actions[0].data ? payload.actions[0].data : {};
+    let pmDetails = getPmIdAndPmTabFromURL(actionPayload.iframe_url);
     return {
         name: payload.name,
         actionType: actionPayload.type,
         choiceId: String(actionPayload.choice_id),
         requestFromPm: false,
-        pmId: getPmIdFromURL(actionPayload.iframe_url),
+        pmId: pmDetails[0],
```
leodoin, line 22: here we could use:
{
...
  pmId:pmDetails.pmId,
  pmTab:pmDetails.pmTab,
...
}

leodoin (changes requested):
Great job!! 

A few notes:
- have you checked if the JS tests for the JSReceiver are still passing? Either away it'd be important to add this pmTab to those tests as well before merging
- most of my suggestions on this review are rather for better code readability, from a implementation perspective you code is rock solid!

cmplibrary/src/main/res/raw/js_receiver.js
Suryakantj: Good suggestion. Done.

cmplibrary/src/main/res/raw/js_receiver.js
Suryakantj: Done

cmplibrary/src/main/res/raw/js_receiver.js
Suryakantj: Good suggestion, Done.

Suryakantj:
> Great job!!
> 
> A few notes:
> 
> * have you checked if the JS tests for the JSReceiver are still passing? Either away it'd be important to add this pmTab to those tests as well before merging
> * most of my suggestions on this review are rather for better code readability, from a implementation perspective you code is rock solid!

Thank you @leodoin , I have made changes as suggested for javascript code and tests

PR #275, opened 2020-10-14
[SP-5105] Add isMessageVisible check before loading SDK

Suryakantj:


app/src/main/java/com/sourcepoint/example_app/MainActivity.java
```diff
@@ -24,6 +24,8 @@
 
     private PropertyConfig config;
 
+    private boolean isMessageVisible = false;
```
AAverin, line 27: This will not work if user have already accepted Native Message once. Assuming you create NativeMessage in `onResume()` or `onStart()`, the flag will always be false because consent will no longer be actually displayed on the screen, but NativeMessage would still be created every time.

leodoin (commented):
It has been tested on our exampleapp -> on the GC event the CustomNativeMessage is dumped.

andresilveirah:
We decided not to merge this PR. Although it does improve on the fact that there'll be only one instance of the ConsentLib on the Heap, as soon as the GC runs, all stale instances will be collected and the memory will be free.

PR #277, opened 2020-10-23
[SP-5183] Add OTT PM support

Suryakantj:
- Now we have dedicated 'privacy manager' for OTT environment.

- OTT PM will act as first layer message

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
+++ b/cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
@@ -26,6 +26,7 @@ public class GDPRConsentLib {
     private final String pmId;
 
     private final String PM_BASE_URL = "https://notice.sp-prod.net/privacy-manager/index.html";
+    private final String OTT_PM_BASE_URL = "https://notice.sp-prod.net/privacy-manager-ott/index.html";
     private final onBeforeSendingConsent onBeforeSendingConsent;
     private final OnNoIntentActivitiesFound onNoIntentActivitiesFound;
 
@@ -56,6 +57,7 @@ public class GDPRConsentLib {
     final messageFinishedCallback messageFinished;
     final onActionCallback onAction;
     final boolean shouldCleanConsentOnError;
+    final boolean isOTT;
 
     public boolean isNative, isPmOn = false;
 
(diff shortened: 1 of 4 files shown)


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -599,7 +602,8 @@ String pmUrl(String privacyManagerId, String pmTab) {
         params.add("pmTab="+pmTab);
         params.add("site_id="+ propertyId);
         if (consentUUID != null) params.add("consentUUID=" + consentUUID);
-        return PM_BASE_URL + "?" + TextUtils.join("&", params);
+        String PM_URL = isOTT ? OTT_PM_BASE_URL : PM_BASE_URL;
+        return PM_URL + "?" + TextUtils.join("&", params);
```
andresilveirah, line 606: It'd be cool to test that if the `isOTT` flag is set, the WebView gets loaded with the correct URL.

Suryakantj: @andresilveirah I have added unit test in ConsentLibBuilder class to check isOTT flag set

andresilveirah: @Suryakantj right, you added a spec for the setter. I'm referring to the actual logic of changing from one PM url to another based on that flag.

Suryakantj: @andresilveirah As suggested I have added unit to check correct pmURL.

cmplibrary/src/test/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLibTest.java
```diff
@@ -205,4 +206,11 @@ public void customConsentTo() throws ConsentLibException, JSONException {
         assertEquals("foo_category", requestParams.getJSONArray("categories").get(0));
         assertEquals("foo_legIntCategory", requestParams.getJSONArray("legIntCategories").get(0));
     }
+
```
andresilveirah, line 209: nice!

andresilveirah (approved):

PR #280, opened 2020-11-07
Capture webview build crash and test it

leodoin:


cmplibrary/src/main/java/com/sourcepoint/gdpr_cmplibrary/GDPRConsentLib.java
```diff
@@ -169,17 +168,20 @@ public static ConsentLibBuilder newBuilder(Integer accountId, String property, I
...
         mCountDownTimer = b.getTimer(onCountdownFinished());
-
         sourcePoint = b.getSourcePointClient();
-
         storeClient = b.getStoreClient();
+
+        // :warning: following methods depend on storeClient initialization
         setConsentData(b.authId);
+        try {
+            webView = buildWebView(b.getContext());
```
andresilveirah, line 181: @leodoin what if we used the fix you implemented but moved the instantiation to the webview to a separate method? Something like:
```java
ConsentWebView getWebView() {
  return webview == null ? buildWebView(context) : webview;
}
```

The downside of this approach is that we'd have to keep a reference to the `context`. But that object is already in memory and it's not going to be released throughout the whole app lifecycle. Meanwhile the `webview` is not used the majority of the time (users already gave consent). 

What do you think?

leodoin:
I think that what we fain from it is negligible,  since the time for
instantiating the webview is minimal - what is costly is the time the load
it with the consent layer resources from the url. Keeping tge reference to
the context is possibly not harmful,  but even better would be to keeping
no references at all (:
Let's align on a call?

andresilveirah:
@leodoin yeah, I think we won't gain "time performance" but rather lower our memory footprint. I'm not sure of the exact number, but I think most of the `40mb` the SDK uses in memory should be coming from the `webview` being instantiated.

leodoin:
> @leodoin yeah, I think we won't gain "time performance" but rather lower our memory footprint. I'm not sure of the exact number, but I think most of the `40mb` the SDK uses in memory should be coming from the `webview` being instantiated.

Very good point sir. I will branch out from this one for this reactor and keep both branches so we can profile them and take an informed decision. Keep tuned!

ciriti:
Hi all!

Where can I find more info about this MR? Is there any Jira ticket I can refer to?

Suryakantj (approved):

PR #306, opened 2020-12-28
[SP-5912] Automate SDK release

carmelo-iriti:


carmelo-iriti:
It was merged using the release/0.0.20 branch where I added some missing variable

PR #311, opened 2021-01-04
[SP-5910]Select privacy manager tab

Suryakantj:


andresilveirah:
@carmelo-iriti I believe we should change the implementation of this feature.
Instead of setting a property on the `ConsentLib` (and its builder) , the idea is to add a second `showPm` method that accepts a `PrivacyManagerTab` as parameter. 

Something like:

```java
void show(PrivacyManagerTab tab) {
  // load pm here 
}
```

carmelo-iriti:
Sure, it sounds good

andresilveirah:
@carmelo-iriti , @vilasma brought up a valid point regarding the PM tab. If the PM tab is only used in the `showPM` method, then it won't affect the PM when coming from the message.

carmelo-iriti (approved):

PR #312, opened 2021-01-05
[SP-5903] Test matrix GitHub workflow

Suryakantj:


carmelo-iriti:
work not compatible with v6

PR #321, opened 2021-02-19
Make MessageLanguage getter usable

avalanchas:
- fixes an appearant typo in the message getter for the name
- name is now the language, so that it is usable for API consumers


avalanchas:
Guys your test pipeline is unfortunately broken 😢 can you please check out this request manually?

carmelo-iriti:
Changes not compatible with v6 version

Issue #322, opened 2021-02-27
Feature request: add ability to override the OkHttpClient()


hnljp:
**Describe the feature request**
It is at this moment not possible to override OkHttpClient().

Personally I see two issues with not being able to override the OkHttpClient()
1. It is recommended to use a single instance of the OkHttpClient() : https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html
2. We use our own custom OkHttpClient() to fix issues certificate issues, on API 21. It would be nice, to be able also to use this for this library.

carmelo-iriti:
HI @hnljp ,

this change requires developing time, what issues is triggering your request?

carmelo-iriti:
Since I didn’t get any answer, I'm closing this ticket

PR #323, opened 2021-02-27
#322 : Add option to specify the used OkHttpClient().

hnljp:
See Issue #322

### DIFF
--- a/cmplibrary/src/test/java/com/sourcepoint/gdpr_cmplibrary/ConsentLibBuilderTest.java
+++ b/cmplibrary/src/test/java/com/sourcepoint/gdpr_cmplibrary/ConsentLibBuilderTest.java
@@ -13,6 +13,8 @@ import org.robolectric.RobolectricTestRunner;
 
 import java.lang.reflect.Field;
 
+import okhttp3.OkHttpClient;
+
 import static junit.framework.TestCase.assertTrue;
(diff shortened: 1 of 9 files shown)


cmplibrary/src/test/java/com/sourcepoint/gdpr_cmplibrary/ConsentLibBuilderTest.java
```diff
@@ -83,7 +85,7 @@ private Field getDeclaredFieldAccess(String fieldName) throws Exception {
 
     @Test
     public void ConsentLibBuilder(){
```
lyndseyBeil, line 87: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

carmelo-iriti:
Changes not compatible with v6 version

PR #427, opened 2021-10-11
[SP-7926] US Privacy String Set to "" when no message fires

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/exposed/SPConsents.kt
```diff
@@ -21,7 +21,7 @@ data class SPCCPAConsent(
     val consent: CCPAConsent
 )
 
-data class Grant(
+data class GDPRPurposeGrants(
```
carmelo-iriti, line 24: renamed to make it closer to IOS

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdpr.kt
```diff
@@ -231,9 +233,17 @@ private class DataStorageGdprImpl(context: Context) : DataStorageGdpr {
     }
 
     override fun clearGdprConsent() {
+
+        val spEditor = preference.edit()
+        preference
+            .all
+            .filter { it.key.startsWith("IABTCF_") }
```
andresilveirah, line 240: Here you could use the constant you defined above `IABTCF_KEY_PREFIX`.

carmelo-iriti: true

PR #459, opened 2022-01-19
[SP-8568] Adding a default case with empty string in the MessageLanguage enum

carmelo-iriti:


carmelo-iriti:
This MR was linked to the ticket [SP-8427](https://sourcepoint.atlassian.net/browse/SP-8427) which was closed. 
The current MR can be discarded.

PR #460, opened 2022-01-20
Clarify OTT PM surface and resurface in README.md

tom-win87:
Clarify OTT PM surface and resurface


tom-win87:
Was asked to clarify how an OTT privacy manager is surfaced for the first time (loadmessage) and how an OTT privacy manager is resurfaced (loadOTTPrivacyManager)

PR #484, opened 2022-04-11
[SP-8824] Add support of groupPmId and childPmId

carmelo-iriti:


samples/metaapp/src/main/sqldelight/com.sourcepointmeta.metaapp.db/11.sqm
```diff
@@ -24,7 +24,7 @@
 -- COMMIT;
```
andresilveirah, line 24: github is showing we moved the file  `.../10.sqm` to `.../11.sqm`. Did you mean to copy the file or to move? Or is it just GitHub showing something wrong?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/exposed/SpConfig.kt
```diff
@@ -17,8 +17,14 @@ data class SpConfig(
...
     @JvmField val campaignType: CampaignType,
-    @JvmField internal val targetingParams: List<TargetingParam>
-)
+    @JvmField internal val targetingParams: List<TargetingParam>,
+    @JvmField val groupPmId: Long? = null
+) {
+    constructor(
+        campaignType: CampaignType,
+        targetingParams: List<TargetingParam>
+    ) : this(campaignType, targetingParams, null)
```
andresilveirah, line 26: Should `groupPmId` be part of the `constructor`?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -135,23 +147,40 @@ private class CampaignManagerImpl(
     fun getGdprPmConfig(pmId: String?, pmTab: PMTab): Either<PmUrlConfig> = check {
         val uuid = dataStorage.getGdprConsentUuid()
         val siteId = dataStorage.getPropertyId().toString()
+        val childPmId: String? = dataStorage.gdprChildPmId
+
+        val usedPmId = childPmId ?: pmId
```
andresilveirah, line 152: Is this method used when the dev call `loadGDPRPrivacyManager`? If it is, we should always use the `pmId` the dev pass as parameter. 
The `childPmId` should only be used if the dev call the new method `loadGDPRGroupPrivacyManager` (same for CCPA).

samples/metaapp/src/main/sqldelight/com.sourcepointmeta.metaapp.db/11.sqm
carmelo-iriti: this is the migration of SQLDelight library used when you add a new field in a table

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/exposed/SpConfig.kt
carmelo-iriti: The `groupId` tells us that the property has a group of campaigns.
This value should be bound to the property rather to the campaigns

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: This is difficult, who need to have the priority? 
Should the user's pmId override the childPmId coming from the server or the opposite? (Right now I gave priority to the server)
What if the user keep using the pmId from the group and he gets the pm with the placeholder?

samples/metaapp/src/main/sqldelight/com.sourcepointmeta.metaapp.db/11.sqm
andresilveirah: Got it. But did it create a new migration file or it moved the previous one to the new?
GitHub shows it renamed the file `10.sqm` to `11.sqm`.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/exposed/SpConfig.kt
andresilveirah: Right, that makes sense. The group pm id is bound to property and the child pm id is a campaign attribute (which the backend return in the UserConsent).

samples/metaapp/src/main/sqldelight/com.sourcepointmeta.metaapp.db/11.sqm
carmelo-iriti: In this case creating a new one or renaming the previous one, doesn’t matter

Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -78,7 +78,11 @@ private class ServiceImpl(
                 nc.sendConsent(it, env, consentActionImpl)
             }
             .executeOnRight {
-                dataStorage.saveLocalState(it.localState)
+                dataStorage.run {
+                    saveLocalState(it.localState)
+                    println("savedConsent = true")
```
Nevazhnovu, line 83: @carmelo-iriti is this `println` desirable?

Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
carmelo-iriti: Right , let me remove it

Nevazhnovu (approved):

PR #492, opened 2022-05-09
[DIA-273] Merge the loadPrivacyManager and loadOTTPrivacyManager functions

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -297,6 +297,8 @@ internal class SpConsentLibImpl(
 
         val gdprGroupPmId = campaignManager.getGroupId(campaignType)
 
+        pLogger.i("loadPm - isOtt: ", isOtt.toString())
```
Nevazhnovu, line 300: Is this log desirable? Just asking 👀

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -262,6 +263,13 @@ private class CampaignManagerImpl(
             ?: false
     }
 
+    override fun isCampaignOtt(campaignType: CampaignType): Boolean {
+        return when (campaignType) {
+            CampaignType.GDPR -> dataStorage.isGdprOtt
+            CampaignType.CCPA -> false
```
Nevazhnovu, line 269: Does it mean we don't have OTT for CCPA at all ?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
carmelo-iriti: yes, pLogger in prod is an empty object, in debug allows us to test if the `isOtt ` has the right value

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
Nevazhnovu: I'm not aware either

PR #495, opened 2022-05-19
[DIA-579] `List<Any>` is substituted with `List<String>` in `CCPAConsentInternal` class

Nevazhnovu:
The type changed for the `rejectedCategories` and `rejectedVendors`


carmelo-iriti:
It looks good 👍 
We should also add a test for it into the `ConsentRespExtKtTest` file. Take as example the test at the line 104, if you need and input, you can use the Metaapp

carmelo-iriti (approved):
well done! 👍

PR #508, opened 2022-07-15
[DIA-547] Adding docs about SpUtil in the README file

carmelo-iriti:


andresilveirah (approved):
Really good and nice catch removing the `README` file from our github workflow!

PR #534, opened 2022-09-12
[DIA-1165] Messages api

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/v7/MessagesApiModel.kt
```diff
@@ -0,0 +1,69 @@
...
+    abstract val thisContent: JSONObject
+    abstract val type: String
+    abstract val dateCreated: String?
+    abstract val url: HttpUrl?
+    abstract val message: JSONObject?
+    abstract val messageMetaData: JSONObject?
+    abstract val messageSubCategory: MessageSubCategory
+}
+
+internal data class GdprMess(
```
andresilveirah, line 27: Do you mind renaming it to `GdprMessage` instead of "mess" 😅 (I understand it's just me being picky). Same for `CcpaMess`.

andresilveirah (approved):

PR #541, opened 2022-10-18
[DIA-1167] implement calls to consent service

Nevazhnovu:


ui-test-util/jsonFiles/v7/choice_all.json
```diff
@@ -0,0 +1,261 @@
+{
+  "ccpa":{
+    "signedLspa":false,
+    "dateCreated":"2022-10-16T12:57:57.264Z",
+    "cookies":[
```
andresilveirah, line 5: @Nevazhnovu we should ignore `signedLspa`, `cookies`, `actions` from both `ccpa` and `gdpr` objects. It's not used in the SDKs so less code to look at :)

carmelo-iriti:
@Nevazhnovu we switched from this custom way of parse the json to _kotlinx-serialization_, this means that we have to wait the message flow MR and then rewrite this story. Please ask me before starting working on story.

PR #542, opened 2022-10-19
[DIA-1368] sp.renderingAppError errors are not being listened by the SDK

carmelo-iriti:


carmelo-iriti:
@andresilveirah this is an example of what we'll deliver to the client 
`{"name":"sp.renderingAppError","messageId":"525881","fromPM":false,"invalidSignature":true}`

carmelo-iriti:
@andresilveirah yes, we have, the exception is `RenderingAppException` but we don’t have the code to send up to the server

carmelo-iriti:
@andresilveirah on ios you use sp_metric_rendering_app_error, I'm adding it for this exception

PR #557, opened 2022-11-28
[DIA-1112] Update uspstring on CCPA

Nevazhnovu:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -90,6 +90,7 @@ private class ServiceImpl(
                     CCPA -> {
                         dataStorage.saveCcpaConsentResp(it.userConsent ?: "")
                         dataStorage.saveCcpaConsentUuid(it.uuid)
+                        dataStorage.saveUsPrivacyString(it.content.getJSONObject("userConsent").getString("uspstring")?:"")
```
carmelo-iriti, line 93: If the `userConsent` is not there, the client application will crash directly, please add some check before using the `getJSONObject` api

carmelo-iriti (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
Nevazhnovu: @carmelo-iriti done, check it out!

carmelo-iriti: We are getting there 👍🏻
1. Let’s move the extensions in the `ConsentRespExt.kt` because they are part of the consent not of the Service
2. Inside the `ConsentRespExtKtTest.kt` we need 2 tests, 
      a. one will use a JSONObject with the following content and the Ext check that it gets `1---` 
`{
  "consents": {
    "USPString": "1---"
  }
}`
   b. one will use an empty JSONObject() and the Ext check that it gets `null`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/ConsentRespExt.kt
```diff
@@ -130,3 +131,23 @@ internal fun Map<String, Map<String, Boolean>>.toAcceptedCategories(): Iterable<
     val falseCategories = partitions.second.map { it.first }.toSet()
     return trueCategories.minus(falseCategories)
 }
+
+fun JSONObject.getStringOrNullByKey(
```
carmelo-iriti, line 135: Using a `try{ .. } catch(e){ return null }` is an anti-pattern in java, use the `isNull` api instead

carmelo-iriti (changes requested):

carmelo-iriti:
This story should be merged in the `v6/develop` branch because `develop` is going to host the new v7

carmelo-iriti (approved):

PR #565, opened 2023-01-11
[DIA-1601] Allow OTT SDKs to surface OTT Privacy Managers without active campaign

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLib.kt
```diff
@@ -59,6 +59,8 @@ interface SpConsentLib {
         successCallback: CustomConsentClient
     )
 
+    @Deprecated("loadOTTPrivacyManager will be removed in the next releases.")
```
carmelo-iriti, line 62: @andresilveirah does this message work?

PR #567, opened 2023-01-24
[DIA-1708] Create a new ott enum type to be used in the loadPrivacyManager api

carmelo-iriti:


README.md
```diff
@@ -286,6 +286,19 @@ Java
     spConsentLib.loadPrivacyManager("<PM_ID>", CampaignType.CCPA); // For a CCPA campaign
 ```
 
+Another option to use to **resurface** the privacy manager, is the following:
```
carmelo-iriti, line 289: @andresilveirah

andresilveirah: @carmelo-iriti if the method `loadPrivacyManager` is generic (ie. it works for all types of PM) why are we limiting its 3rd parameter to an OTT type?

What if I want to load a "app" Privacy manager?

carmelo-iriti: we need to add in the `OttPmType` an APP item or we can delete the `generic ` word in the description

PR #574, opened 2023-02-01
[DIA-1742] Fixed empty consent in an edge case

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModelExt.kt
```diff
@@ -18,9 +21,9 @@ internal fun GdprCS.toGDPRUserConsent(): GDPRConsentInternal {
         tcData = TCData?.toMapOfAny() ?: emptyMap(),
         grants = grants ?: emptyMap(),
         euconsent = euconsent ?: "",
-        acceptedCategories = acceptedCategories ?: emptyList(),
+        acceptedCategories = acceptedCategories,
         childPmId = null,
-        applies = gdprApplies ?: applies ?: false,
+        applies = gdprApplies ?: applies ?: TCData?.fromTcDataToGdprApplies(),
```
andresilveirah, line 26: @carmelo-iriti this we'll probably delete since we'll ignore all `applies` except for the one coming from `/meta-data`, right?

carmelo-iriti: since the `apply` from '/meta' is directly connected to the `IABTCF_gdprApplies ` of TCData, it would make sense to get that value from here with the 'fromTcDataToGdprApplies' function. I choose this approach because right now taking the apply from the SP needs to many changes

andresilveirah: The downside of getting it from `TCData` is that it forces us to know about the contents of `TCData`.
But that's ok if this is just a temporary approach until we implement the other logic to get the applies from `/meta-data`

carmelo-iriti: Correct me if I’m wrong but TCData must always be there if the campaign is a GDPR one

andresilveirah: That's correct. `TCData` _should_ always be present if the campaign is GDPR. But we have no guarantee over its contents. It's dictated by the TCF spec (which can change). That's why I suggest we use the `applies` from `/meta-data` instead.

PR #590, opened 2023-02-17
[DIA-1735] linked no action and unknown values added

Nevazhnovu:


carmelo-iriti:
it looks that this is the same MR of https://github.com/SourcePointUSA/android-cmp-app/pull/591/files

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/ConsentRespExt.kt
```diff
@@ -73,9 +73,10 @@ internal fun Map<String, Any?>.toCCPAUserConsent(uuid: String?, applies: Boolean
 
     val status: CcpaStatus = getFieldValue<String>("status")
         ?.let { s ->
-            CcpaStatus.values().find {
+            val status = CcpaStatus.values().find {
                 it.name == s
             }
+            return@let status ?: CcpaStatus.unknown
         }
         ?: fail("CCPAStatus cannot be null!!!")
```
carmelo-iriti, line 81: this condition should be removed

carmelo-iriti (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/ConsentRespExt.kt
```diff
@@ -76,9 +76,9 @@ internal fun Map<String, Any?>.toCCPAUserConsent(uuid: String?, applies: Boolean
             val status = CcpaStatus.values().find {
                 it.name == s
             }
-            return@let status ?: CcpaStatus.unknown
+            status ?: CcpaStatus.unknown
```
carmelo-iriti, line 79: Let’s also verify if this extension is currently used in the code and if the functions that use it are also used. If they are belonging to the v6 and are not used in the v7 we can delete them

carmelo-iriti (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/ConsentRespExt.kt
Nevazhnovu: @carmelo-iriti this function is not used in v7 at all. It is used in v6 only.

carmelo-iriti: In this case I would say that we can remove it and its all its flow

Nevazhnovu: Wait a minute. It is used while changing `localStateStatus`

PR #627, opened 2023-04-27
[DIA-1981] Write a more meaningful message for the timeout exception

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exception/Exceptions.kt
```diff
@@ -227,7 +227,7 @@ internal class InvalidArgumentException @JvmOverloads constructor(
  */
 internal class ConnectionTimeoutException @JvmOverloads constructor(
     cause: Throwable? = null,
-    description: String,
+    description: String = "A timeout has occurred when requesting the message data. You can extend the timeout using the messageTimeout config parameter.",
```
andresilveirah, line 230: @carmelo-iriti is this the timeout to the `/messages` request or is it the timeout to the "message flow"?

PR #630, opened 2023-05-03
[DIA-2134] Trigger the timeout for RenderingApp from the onConsentReady callback

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/SPWebViewClient.kt
```diff
@@ -30,14 +30,13 @@ internal class SPWebViewClient(
     override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
         super.onPageStarted(view, url, favicon)
         timer.executeDelay(messageTimeout) {
-            onError(ConnectionTimeoutException(description = "A timeout has occurred when loading the message"))
+            onError(RenderingAppConnectionTimeoutException(description = "A timeout has occurred when loading the consent message"))
```
andresilveirah, line 33: @carmelo-iriti what do you think about changing the message to something like:
```
There was an error while loading the rendering app. onConsentReady was not called within X seconds.
```
Replacing X by the timeout amount.

carmelo-iriti: done

PR #632, opened 2023-05-04
[DIA-2121] Added ViewManagerException

Nevazhnovu:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exception/ExceptionCodes.kt
```diff
@@ -14,6 +14,7 @@ internal object CodeList {
     val INVALID_ON_ACTION_EVENT_PAYLOAD = ExceptionCodes("sp_metric_invalid_onAction_event_payload")
     val URL_LOADING_ERROR = ExceptionCodes("sp_metric_url_loading_error")
     val WEB_VIEW_ERROR = ExceptionCodes("sp_metric_web_view_error")
+    val VIEW_MANAGER_ERROR = ExceptionCodes("sp_metric_view_manager_error")
```
carmelo-iriti, line 17: The ViewManager is a generic object that takes care of a piece of the app, using `sp_metric_view_manager_error` doesn’t make us understand the real issue which is an error related to the impossibility of creating e WebView to visualise the message.

I would change the reference to the viewManager with something more specific like:
`sp_metric_webview_creation_error`

carmelo-iriti (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exception/ExceptionCodes.kt
carmelo-iriti: Please edit also the names that have the `viewmanager` word

PR #635, opened 2023-05-09
[DIA-2106] Make UUID public on CCPAConsent

bohdan-go-wombat:
**JIRA Ticket** - https://sourcepoint.atlassian.net/browse/DIA-2106

**Description**
This PR introducing exposure of UUID for CCPAConsent object, which leads to our users be able to use UUID of CCPA.


Nevazhnovu (commented):
@bohdan-go-wombat as you may notice, the ui-test-app test suite is not passing. Some tests are flaky and that's okay behavior. Launching all the test suites locally before creating the PR would be preferable. You can relaunch it in GitHub to see if it passes next time.

carmelo-iriti (approved):

PR #643, opened 2023-05-25
[DIA-2198] Manipulate CCPA String

bohdan-go-wombat:


cmplibrary/src/test/java/com/sourcepoint/cmplibrary/util/extensions/CcpaCSExtKtTest.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Test case that covers the case when:
+     * applies == null
+     * then the consent string should be == "1---"
+     */
+    @Test
+    fun `generateConsentString - WHEN called with applies as null THEN should return default CCPA consent string`() {
+        // GIVEN
+        val ccpaCS = CcpaCS(
```
andresilveirah: Creating a `CcpaCS` seems a bit cumbersome. Since the only thing that we care (in this test) is changing the values of `signedLspa`, `applies` and the consent status, you could create a helper method like this:
```kotlin
fun genCCPA(signedLspa: Boolean, applies: Boolean, status: CCPAStatus): String {
    return CcpaCS(
            actions = null,
            applies = applies,
            ccpaApplies = null,
            consentedAll = null,
            cookies = null,
            dateCreated = null,
            gpcEnabled = null,
            newUser = null,
            rejectedAll = null,
            rejectedCategories = null,
            rejectedVendors = null,
            signedLspa = signedLspa,
            status = status,
            uspstring = "",
            uuid = null,
        )
}
```

And use the method on each test case.

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/util/extensions/CcpaCSExtKtTest.kt
```diff
@@ -0,0 +1,185 @@
...
+            uspstring = null,
+            uuid = null,
+        )
+        val expected = "1---"
+
+        // WHEN
+        val actual = ccpaCS.generateConsentString()
+
+        // THEN
+        actual.assertEquals(expected)
```
andresilveirah: For brevity, we could substitute:
```kotlin
val expected = "1---"

// WHEN
val actual = ccpaCS.generateConsentString()

// THEN
actual.assertEquals(expected)
```
with
```kotlin
ccpaCS.generateConsentString().assertEquals("1---")
```

In my opinion it keeps things shorter and still easy to understand what's going on (specially if we use that helper method I mentioned above)

Nevazhnovu:
@bohdan-go-wombat good to know how `CCPAExt.kt` is implemented, thank you for the ping!

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/util/extensions/CcpaCSExtKtTest.kt
bohdan-go-wombat: I changed the place for `generateCcpaUspString()` method, so I changed the place for test suite as well, created this helper method you've mentioned, so now from my pov it looks clean

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/util/extensions/CcpaCSExtKtTest.kt
bohdan-go-wombat: from my pov, if we take an abstact term of what is the purpose of the test, then it is to match the `actual` result to the `expected` one

in this case, I also wanted to make sure that `generated` result right from the helper class `SpUtils` generates the same value, as we get from the `actual`

that is why, from my pov, the original way is more readable, because you can see what are `actual` and `expected` results

PR #657, opened 2023-06-27
[DIA-2151] Removed singleShotPM from ConsentActionImpl

Nevazhnovu:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ClientEventManager.kt
```diff
@@ -61,7 +61,7 @@ private class ClientEventManagerImpl(
                 ActionType.CUSTOM,
                 ActionType.MSG_CANCEL,
                 ActionType.PM_DISMISS -> {
-                    if (!action.requestFromPm || action.singleShotPM) {
+                    if (!action.requestFromPm) {
```
bohdan-go-wombat, line 64: I think we should move away from using `!` and use `.not()` instead, since it is more of a kotlinish way

bohdan-go-wombat (approved):

andresilveirah:
@Nevazhnovu there are other places where `singleShot` is being used. Did you consider removing them as as well? Also, did you check if the callbacks behave correctly for a campaign where a PM is used as 1st layer message?

Nevazhnovu:
@andresilveirah it's used inside `js_reciever.js` (to communicate with backend I believe) and nowhere else in the Android SDK itself, after I removed the entrances of `singleShot` in this PR.
 I don't think that's a good idea to remove it completely, just in the case described in the ticket (which I did)

Nevazhnovu:
I've removed it completely and manually tested with 3 properties - it works great

andresilveirah:
@Nevazhnovu I noticed `onSPFinished` was not being called if the user would load the PM (not via 1st layer message) and dismiss it. I took the liberty to fix that, refactoring `ClientEventManager` in the process and written a few tests for it.

andresilveirah (approved):

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/consent/ClientEventManagerTest.kt
```diff
@@ -38,118 +35,116 @@ class ClientEventManagerTest {
             spClient = spClient,
             executor = MockExecutorManager()
         )
-    }
-
-    @Test
-    fun `GIVEN 2 successfully sendConsent (GDPR, CCPA) calls, TRIGGER 1 onSpFinish`() {
 
         every { consentManagerUtils.gdprConsentOptimized }.returns(Either.Right(GDPRConsentInternal()))
         every { consentManagerUtils.ccpaConsentOptimized }.returns(Either.Right(CCPAConsentInternal()))
```
bohdan-go-wombat, line 40: why did you decide to put those 2 lines into `setup`? do we use those lines in the majority of the test cases?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ClientEventManager.kt
```diff
@@ -34,42 +31,35 @@ private class ClientEventManagerImpl(
...
                 }
                 ActionType.CUSTOM,
                 ActionType.MSG_CANCEL,
                 ActionType.PM_DISMISS -> {
-                    if (!action.requestFromPm || action.singleShotPM) {
-                        if (cNumber > 0) cNumber--
+                    if (isFirstLayerMessage) {
+                        campaignsToProcess--
                     }
+                    isFirstLayerMessage = true
```
bohdan-go-wombat, line 53: why if we dismiss the PM we set `isFirstLayerMessage ` as `true` in this case?

andresilveirah: There are 3 cases in which a user find themselves in the PM.
1. the user landed in the Privacy Manager via the 1st layer message. In this case, dismissing the PM should should return the user to the first layer message, thus we should reset the value of `isFirstLayerMessage`
2. the PM _is_ the first layer message, in this case, dismissing the PM should close the message entirely, here we reset `isFirstLayerMessage` just to have that variable at its "original" state of `true`.
3. the PM is loaded on demand (ie. via `loadPrivacyManager` methods). Same as the item before, simply reseting the value to `true` (even though its value should be `true` anyways).

PR #658, opened 2023-07-04
[DIA-2198] Manipulate CCPA consent string based on applies

bohdan-go-wombat:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -540,31 +542,50 @@ private class CampaignManagerImpl(
...
-                sampleRate?.let { i ->
-                    if (i != dataStorage.ccpaSamplingValue) {
-                        dataStorage.ccpaSamplingValue = i
-                        dataStorage.ccpaSamplingResult = null
-                    }
+        get() = dataStorage.gdprConsentStatus != null || dataStorage.uspstring.isNullOrEmpty().not()
+
+    override fun handleMetaDataResponse(response: MetaDataResp?) {
+        // update meta data response in the data storage
+        metaDataResp = response
```
andresilveirah, line 549: @bohdan-go-wombat maybe this comment doesn't belong to this PR, but do we need to store the responses? 

If you think it makes sense, we could remove `metaDataResp` from the local storage already.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -540,31 +542,50 @@ private class CampaignManagerImpl(
...
+
+        if (response == null) return
+
+        // handle ccpa
+        response.ccpa?.apply {
+            applies?.let { i ->
```
andresilveirah, line 555: @bohdan-go-wombat could you rename `i` to something more intuitive in the lines below? Otherwise I don't see the improvement using `i` over `it`.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -540,31 +542,50 @@ private class CampaignManagerImpl(
...
+                if (i != dataStorage.ccpaSamplingValue) {
+                    dataStorage.ccpaSamplingValue = i
+                    dataStorage.ccpaSamplingResult = null
+                }
+            }
+        }
+
+        // handle gdpr
+        response.gdpr?.apply {
+            applies?.let { i -> dataStorage.gdprApplies = i }
```
andresilveirah, line 573: I thought we'd get rid of `dataStorage.gdprApplies` and deal only with the consent object `.applies` attribute.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorage.kt
```diff
@@ -35,6 +35,7 @@ internal interface DataStorage : DataStorageGdpr, DataStorageCcpa {
 
     var consentStatusResponse: String?
     var gdprConsentStatus: String?
+    val ccpaApplies: Boolean
```
andresilveirah, line 38: Why did you decide to move `ccpaApplies` from `DataStorageCcpa` class here?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageImpl.kt
```diff
@@ -79,6 +85,18 @@ private class DataStorageImpl(
                 .apply()
         }
 
+    /**
+     * Parameter that accesses applies value from ccpaConsentStatus directly
+     *
+     * By default - false (if there are no ccpaConsentStatus in data storage or applies value is null)
+     */
+    override val ccpaApplies: Boolean
```
andresilveirah, line 93: @bohdan-go-wombat what about the `set`? If I understood it right, we access `ccpaApplies` by reaching out to the ccpa consent object, but can still set its value independently. Is that right?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorage.kt
bohdan-go-wombat: @andresilveirah because it is a "mirroring" parameter, and we decided to actually access it via `ccpaConsentStatus` from the DataStorageCcpa

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
bohdan-go-wombat: @andresilveirah agree

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
bohdan-go-wombat: @andresilveirah I think comments like that could actually help a lot with a readability of the code for the future refferences

regarding `metaDataResp`, I will take a look, but as I already mentioned, I strongly not recommend do a lot of refactoring in the PR, that is responsible for a speciffic fix/feature

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
bohdan-go-wombat: @andresilveirah since this is a PR to fix CCPA related issue, I decided to refactor CCPA related code only, and for the GDPR code I can create a ticket to do it separatelly, since:
1. it will help us to merge this task asap
2. it will prevent us from making mistakes in the part of the code, that is not related to the issue we are trying to fix
3. it will prevent us from breaking the purpose of the pull request (which is to fix a problem with uspstring)

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageImpl.kt
bohdan-go-wombat: when I refactored I made `ccpaApplies` `val`, since we no longer need to update the `mirroring` value, and we will just update the local `ccpaConsentStatus` object (the one that is stored in the data store and being used in our app)

bohdan-go-wombat: getter is needed to have the easier access to the `ccpaApplies`, for example rather then calling something like:
`dataStorage.ccpaConsentStatus?.let { verifiedCcpaConsentStatusString ->
            check { JsonConverter.converter.decodeFromString<CcpaCS>(verifiedCcpaConsentStatusString) }
                .getOrNull()
                ?.applies`
we will call:
`dataStorage.ccpaApplies`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
andresilveirah: @bohdan-go-wombat my bad, the comment I was referring to was my GH comment, not the comment on the code 😅 

Regarding not mixing all the refactoring in the same PR, agree with you.

Nevazhnovu (approved):

andresilveirah (approved):

PR #667, opened 2023-07-21
[DIA-2449] Not call onError if pv-data fails

bohdan-go-wombat:
### Requirements
Jira Ticket - [DIA-2449](https://sourcepoint.atlassian.net/browse/DIA-2449)

### Description
This PR introduces a change to dealing with `onError()` callback from the message flow. Whenever there are an error in `postPvData` network call, the SDK **should not** invoke the `onError()` callback, but **should** onvoke all the other in-app metrics.

[DIA-2449]: https://sourcepoint.atlassian.net/browse/DIA-2449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
@@ -212,9 +212,9 @@ internal class SpConsentLibImpl(
                     pLogger.e(
                         "SpConsentLib",
                         """
-                    onError
-                    ${error.message}
-                """.trimIndent()
+                            onError
+                            ${error.message}
+                        """.trimIndent()
                     )
                 }
             },
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
@@ -145,7 +145,7 @@ internal class SpConsentLibImpl(
                 consentManager.sendStoredConsentToClient()
                 clientEventManager.setAction(NativeMessageActionType.GET_MSG_NOT_CALLED)
             },
-            pSuccess = {
+            onSuccess = {
                 val list = it.toCampaignModelList(logger = pLogger)
                 clientEventManager.setCampaignsToProcess(list.size)
                 if (list.isEmpty()) {
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
@@ -192,53 +192,33 @@ internal class SpConsentLibImpl(
                     }
                 }
             },
-            pError = { throwable ->
-                onErrorInMessageFlow(
-                    error = throwable
-                )
-            },
-            onErrorFromPvData = { error, shouldCallOnErrorCallback ->
-                onErrorInMessageFlow(
-                    error = error,
-                    shouldCallOnErrorCallback = shouldCallOnErrorCallback
-                )
-            },
-        )
-    }
-
-    /**
-     * This method handles the error from any network call of a message flow.
-     *
-     * @param error - an error that occurred during the network call
-     * @param shouldCallOnErrorCallback - a flag that decides whether or not to invoke a callback
-     * from the [SpClient] (e.g. business requirements stating that if the error occurs in postPvData()
-     * method then the onError() callback should not be called).
-     */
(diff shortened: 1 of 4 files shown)


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/Service.kt
```diff
@@ -39,7 +39,8 @@ internal interface Service : NetworkClient, CampaignManager {
         messageReq: MessagesParamReq,
         pSuccess: (MessagesResp) -> Unit,
         showConsent: () -> Unit,
-        pError: (Throwable) -> Unit
+        pError: (Throwable) -> Unit,
+        onErrorFromPvData: (Throwable, Boolean) -> Unit,
```
andresilveirah, line 43: What's the advantage of adding a new error callback for `pv-data`? 
Could we simply _not_ call `pError` on `pv-data` and send a call to `custom-metrics` endpoint?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/data/ServiceImplTest.kt
```diff
@@ -189,7 +192,13 @@ class ServiceImplTest {
...
-        sut.getMessages(messageReq = messagesParamReq, showConsent = consentMockV7, pSuccess = successMockV7, pError = errorMock)
+        sut.getMessages(
+            messageReq = messagesParamReq,
+            showConsent = consentMockV7,
+            pSuccess = successMockV7,
+            pError = errorMock,
+            onErrorFromPvData = onErrorFromPvDataMock,
+        )
 
 //        verify(exactly = 0) { errorMock(any()) }
```
Nevazhnovu, line 203: do we still need this comment?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/data/ServiceImplTest.kt
```diff
@@ -210,7 +219,13 @@ class ServiceImplTest {
...
-        sut.getMessages(messageReq = messagesParamReq, showConsent = consentMockV7, pSuccess = successMockV7, pError = errorMock)
+        sut.getMessages(
+            messageReq = messagesParamReq,
+            showConsent = consentMockV7,
+            pSuccess = successMockV7,
+            pError = errorMock,
+            onErrorFromPvData = onErrorFromPvDataMock,
+        )
 
         // TODO
```
Nevazhnovu, line 230: Can the comment be removed?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/Service.kt
Nevazhnovu: @bohdan-go-wombat I might be wrong and taking things out of context, but...
 I would avoid adding new callbacks since it is a major change in a public API. 
 Can we map `PvData` error to the existing `onError` callback?

bohdan-go-wombat: @andresilveirah there are 2 ways of implementing this, one of which the way I implemented it, and the other one is to utilize the already existing `pError` callback with a slight modification, but this option brings a slight confusion, at least for me.

Regarding your suggestion, there are a couple of cons to this:
1. we can't just call a `custom-metrics` endpoint, since we also have a flow to send stored consent to a client via `ConsentManager` and set message error action via `ClientEventManager` in the `pError` callback, so we have to make sure that works as intended as well
2. in order to do so, we would need to add a couple of parameters to a constructor of the `Service` class, and deal with the instances inside this class, which, at least for me, seemed like an overkill

@Nevazhnovu as far as I'm concerned `Service` interface and `ServiceImpl` class are not a part of the public API (if by "public API" you mean the code that will be used by our users directly and not via our SDK API). The `Service` object is being used to create `SpConsentLib` instance, which **is** a part of the public API.

Nevazhnovu: Gotcha, makes sense.

Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -145,7 +145,7 @@ internal class SpConsentLibImpl(
                 consentManager.sendStoredConsentToClient()
                 clientEventManager.setAction(NativeMessageActionType.GET_MSG_NOT_CALLED)
             },
-            pSuccess = {
+            onSuccess = {
```
andresilveirah, line 148: Thanks for this renaming 👍

andresilveirah (approved):

andresilveirah (approved):

PR #669, opened 2023-07-28
[DIA-1961] Extended UnitySpClient for actual Unity v7 port

Nevazhnovu:
These changes help Unity SDK to communicate with native Android SDK

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpClient.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpClient.kt
@@ -45,4 +45,5 @@ interface SpClient {
 
 interface UnitySpClient : SpClient {
     fun onConsentReady(consent: String)
+    fun onSpFinished(consent: String)
 }
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ClientEventManager.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ClientEventManager.kt
@@ -1,6 +1,7 @@
 package com.sourcepoint.cmplibrary.consent
 
 import com.sourcepoint.cmplibrary.SpClient
+import com.sourcepoint.cmplibrary.UnitySpClient
 import com.sourcepoint.cmplibrary.core.ExecutorManager
 import com.sourcepoint.cmplibrary.core.getOrNull
 import com.sourcepoint.cmplibrary.exception.Logger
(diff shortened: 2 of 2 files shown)


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpClient.kt
```diff
@@ -45,4 +45,5 @@ interface SpClient {
 
 interface UnitySpClient : SpClient {
     fun onConsentReady(consent: String)
+    fun onSpFinished(consent: String)
```
andresilveirah, line 48: @bohdan-go-wombat what if we migrate both of these methods to `SpClient`?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ClientEventManager.kt
```diff
@@ -91,6 +92,7 @@ private class ClientEventManagerImpl(
             val spConsentString = spConsent
                 ?.let {
                     spClient.onSpFinished(it)
+                    (spClient as? UnitySpClient)?.onSpFinished(it.toJsonObject().toString())
```
andresilveirah, line 95: So we can always call both `onSpFinished(Object)` and `onSpFinished(String)`, without trying to cast the `spClient` into `UnitySpClient`.

In other words, we're adding support to Unity (and potentially other platforms that deal only with strings) right into our `SpClient` rather than as a separate entity.

bohdan-go-wombat: @andresilveirah I'm not that aware of how things work on the Unity side, but if we merge `UnitySpClient` to `SpClient` (which is the main client for Android SDK users) doesn't that mean that Unity SDK will have to override methods with native Kotlin data classes, such as `View`, `SpConsents`, etc?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpClient.kt
bohdan-go-wombat: @andresilveirah I mean we can, but what are the benefits of that?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ClientEventManager.kt
Nevazhnovu: I would rather keep them separate.

bohdan-go-wombat (approved):

PR #671, opened 2023-07-31
[DIA-2458] Review params to GET /messages endpoint

bohdan-go-wombat:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/messages/OperatingSystemInfoParam.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import android.os.Build
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal data class OperatingSystemInfoParam(
+    @SerialName("name")
+    val name: String? = DEFAULT_ANDROID_OS_NAME,
+    @SerialName("version")
```
andresilveirah, line 11: @bohdan-go-wombat we don't need to add the `@SerialName` notation if the attribute in the data class has the same name as on the final `json`.

andresilveirah: I agree with you, _if_ we'd have many fields with different names, that'd be confusing. But I believe that's not the case right? From what I could see, most `json` fields have identical names as the attributes with a few exceptions. 

We might be getting in the realm of "opinion" now, but I believe this:
```kotlin
@Serializable
data class IncludeData(
    val localState: IncludeDataParam? = null,
    val tcData: IncludeDataParam? = null,
    val campaigns: IncludeDataParam? = null,
    val customVendorsResponse: IncludeDataParam? = null,
    val messageMetaData: IncludeDataParam? = null,
    val webConsentPayload: IncludeDataParam? = null,
)
```
Is clearer than this:
```kotlin
@Serializable
data class IncludeData(
    @SerialName("localState")
    val localState: IncludeDataParam? = null,
    @SerialName("TCData")
    val tcData: IncludeDataParam? = null,
    @SerialName("campaigns")
    val campaigns: IncludeDataParam? = null,
    @SerialName("customVendorsResponse")
    val customVendorsResponse: IncludeDataParam? = null,
    @SerialName("messageMetaData")
    val messageMetaData: IncludeDataParam? = null,
    @SerialName("webConsentPayload")
    val webConsentPayload: IncludeDataParam? = null,
)
```

And, do we need to assign `null` if the attribute is optional? Could we replace, for example, this:
```kotlin
val webConsentPayload: IncludeDataParam? = null,
```
with this:
```kotlin
val webConsentPayload: IncludeDataParam?
```

bohdan-go-wombat: 1. It is definitelly a matter of opinion, and from my point of view this:
``` kotlin
@Serializable
data class IncludeData(
    @SerialName("localState")
    val localState: IncludeDataParam? = null,
    @SerialName("TCData")
    val tcData: IncludeDataParam? = null,
    @SerialName("campaigns")
    val campaigns: IncludeDataParam? = null,
    @SerialName("customVendorsResponse")
    val customVendorsResponse: IncludeDataParam? = null,
    @SerialName("messageMetaData")
    val messageMetaData: IncludeDataParam? = null,
    @SerialName("webConsentPayload")
    val webConsentPayload: IncludeDataParam? = null,
)
```
is more readable and clear than this:
``` kotlin
@Serializable
data class IncludeData(
    val localState: IncludeDataParam? = null,
    @SerialName("TCData") val tcData: IncludeDataParam? = null,
    val campaigns: IncludeDataParam? = null,
    val customVendorsResponse: IncludeDataParam? = null,
    val messageMetaData: IncludeDataParam? = null,
    val webConsentPayload: IncludeDataParam? = null,
)
```
especially, when the IDE highlights what is the parameter of the model itself and what is the name for the serizalization.

---

2. Regarding your second question, we need to have default values, which allows us to create an instance of `IncludeData` without passing `null` to each parameter we don't need, so it looks like this:
``` kotlin
fun generateIncludeDataForMessages(): IncludeData = IncludeData(
    tcData = IncludeDataParam(IncludeDataParamType.RECORD_STRING.type),
    campaigns = IncludeDataParam(IncludeDataParamType.RECORD_STRING.type),
    webConsentPayload = IncludeDataParam(IncludeDataParamType.RECORD_STRING.type),
)
```
and not like this:
``` kotlin
fun generateIncludeDataForMessages(): IncludeData = IncludeData(
    localState = null,
    tcData = IncludeDataParam(IncludeDataParamType.RECORD_STRING.type),
    campaigns = IncludeDataParam(IncludeDataParamType.RECORD_STRING.type),
    customVendorsResponse = null,
    messageMetaData = null,
    webConsentPayload = IncludeDataParam(IncludeDataParamType.RECORD_STRING.type),
)
```

andresilveirah (approved):

PR #672, opened 2023-08-01
[DIA-2460] Review params to GET /choice endpoint

bohdan-go-wombat:


Nevazhnovu (approved):

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/data/network/util/HttpUrlManagerTest.kt
```diff
@@ -233,15 +235,19 @@ class HttpUrlManagerTest {
 
     @Test
     fun `GIVEN a PROD env getChoiceUrl RETURN the prod link`() {
-        val param = ChoiceParamReq(
+        val param = GetChoiceParamReq(
+            choiceType = ChoiceTypeParam.CONSENT_ALL,
             accountId = 22,
+            propertyId = 17801,
```
andresilveirah, line 241: @bohdan-go-wombat can we change the property id to use our "demo" property? The demo property id is `16893`.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/choice/GetChoiceParamReq.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.sourcepoint.cmplibrary.data.network.model.optimized.MetaDataArg
+import com.sourcepoint.cmplibrary.data.network.model.optimized.includeData.IncludeData
+import com.sourcepoint.cmplibrary.data.network.util.Env
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal data class GetChoiceParamReq(
+    val env: Env,
+    val choiceType: ChoiceTypeParam,
+    val metadataArg: MetaDataArg?,
```
andresilveirah, line 12: @bohdan-go-wombat I just realised, `metadataArg` should be called `metadata` and `choiceType` is not part of the query params consumed by the `GET /choice` endpoint. 

Are those used as params or are they present to make your life easier when constructing the `/choice` request?

andresilveirah: To recap, the params should be:
```js
accountId: Int
propertyId: Int
hasCsp: true // hardcoded
metadata: {
  "ccpa"?: { "applies": Boolean },
  "gdpr"?: {  "applies": Boolean }
}
includeCustomVendorsRes: false // hardcoded
withSiteActions: false // hardcoded
includeData: IncludeData
env: "prod" | "localProd" | "stage"
```

andresilveirah: @bohdan-go-wombat can you also review the request to `/messages`? I believe the field `metadata` is called `metadataArgs` there as well.

bohdan-go-wombat: @andresilveirah `GetChoiceParamReq` class is basically an internal wrapper, a class that makes it easier for us to gather all the data for `GET /choice` request.

The assembling of the request happens in the `HttpUrlManager` class in the `getChoiceUrl` method. As you can see, we are adding meta data as a `EncodedQueryParameter` with a key `metadata`:
``` kotlin
override fun getChoiceUrl(param: GetChoiceParamReq): HttpUrl {
        
        val metaData: String? = param.metadataArg?.let { JsonConverter.converter.encodeToString(it) }
        val includeData = JsonConverter.converter.encodeToString(param.includeData)

        return HttpUrl.Builder()
            .scheme("https")
            .host(param.env.host)
            .addPathSegments("wrapper/v2/choice")
            .addPathSegments(param.choiceType.type)
            .addQueryParameter("env", param.env.queryParam)
            .addQueryParameter("accountId", param.accountId.toString())
            .addQueryParameter("propertyId", param.propertyId.toString())
            .addQueryParameter("hasCsp", param.hasCsp.toString())
            .addQueryParameter("withSiteActions", param.withSiteActions.toString())
            .addQueryParameter("includeCustomVendorsRes", param.includeCustomVendorsRes.toString())
            .addEncodedQueryParameter("metadata", metaData)
            .addQueryParameter("includeData", includeData)
            .addQueryParameter("scriptType", scriptType)
            .addQueryParameter("scriptVersion", scriptVersion)
            .build()
    }
```

So in the end, it doesn't matter in this case, how do we name the `metaDataArg` parameter, since we always going to add it up to the request with the `metadata` key.

The same goes with the `choiceType `, since `GetChoiceParamReq` is a wrapper and not a model that is going to be serialized into a request body or something like that (it is a GET request at the end), we are just passing the choice of the user (either `consent-all` or `reject-all`) to build up the request url.

PR #680, opened 2023-08-16
[DIA-2390] Store GPPData in the local storage

bohdan-go-wombat:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -241,15 +245,23 @@ private class ServiceImpl(
...
+
+                            // save tc data in the data storage
+                            it.campaigns?.gdpr?.TCData?.let { tcData ->
+                                dataStorage.tcData = tcData.toMapOfAny()
+                            }
+
+                            // clear out previous and save new gpp data in the data storage
+                            it.campaigns?.ccpa?.gppData?.let { gppData ->
+                                dataStorage.clearGppData()
+                                dataStorage.gppData = gppData.toMapOfAny()
```
andresilveirah, line 258: What you think of adding the `clearGppData()` to the `dataStorage.gppData` setter? This way we don't have to call it always, just by setting `dataStorage.gppData = ...` will clean it for us.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpa.kt
```diff
@@ -127,6 +135,48 @@ private class DataStorageCcpaImpl(context: Context) : DataStorageCcpa {
             }
         }
 
+    override var gppData: Map<String, Any?>
+        get() {
+            val result = TreeMap<String, Any?>()
+            preference.all
+                .filter { it.key.startsWith(DataStorageGdpr.IABTCF_KEY_PREFIX) }
```
andresilveirah, line 142: Here you probably meant to use 
```
KEY_IABGPP_PREFIX
```

Instead of 
```
IABTCF_KEY_PREFIX
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -63,19 +63,35 @@ data class ConsentStatusResp(
...
+    @SerialName("rejectedAll")
+    val rejectedAll: Boolean?,
+    @SerialName("rejectedCategories")
+    val rejectedCategories: List<String>?,
+    @SerialName("rejectedVendors")
+    val rejectedVendors: List<String>?,
+    @SerialName("signedLspa")
+    val signedLspa: Boolean?,
+    @Serializable(with = CcpaStatusSerializer::class)
+    val status: CcpaStatus?,
```
andresilveirah, line 87: `status` doesn't have a `@SerialName("status")` is it intentional?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -88,29 +104,52 @@ data class CcpaCS(
...
+    @SerialName("customVendorsResponse")
+    val customVendorsResponse: CustomVendorsResponse?,
+    @SerialName("dateCreated")
+    val dateCreated: String?,
+    @SerialName("euconsent")
+    val euconsent: String?,
+    @Serializable(with = GrantsSerializer::class)
+    val grants: Map<String, GDPRPurposeGrants>?,
+    @Serializable(with = JsonMapSerializer::class)
+    val TCData: Map<String, JsonElement>?,
```
andresilveirah, line 144: Same for `grants` and `TCData`, missing `@SerialName`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/gpp/dto/GppTernaryType.kt
```diff
@@ -0,0 +1,9 @@
+package com.sourcepoint.cmplibrary.gpp.dto
+
+internal enum class GppTernaryType(
+    val type: String,
+) {
+    YES("yes"),
+    NO("no"),
+    NON_APPLICABLE("na"),
```
andresilveirah, line 8: Could you rename it to `NOT_APPLICABLE`?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpa.kt
bohdan-go-wombat: @andresilveirah clearly noticed, cudos to you, I will also do another round of manual testing of this stuff

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
bohdan-go-wombat: @andresilveirah agree, that makes more sense, since we don't have any case when we update `gppData` without clearing out local data first

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
bohdan-go-wombat: yes, this line was not changed, and the serializer will use the name of the parameter (i.e. `"status"`)

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
bohdan-go-wombat: same here

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
bohdan-go-wombat: @andresilveirah actually, I think we need to clear out data storage outside of the setter, since what if we already have local gpp data, and the response from the BE doesn't have gpp in it? I assume we would have to clear out the local data, since our BE is the single source of truth in this case

andresilveirah: If the response doesn't have GPP data in it, we should set it to `null` (i.e. remove all GPP from local storage). So, having the `clearGppData()` in the setter will work just fine.

bohdan-go-wombat: @andresilveirah you were right, I fixed this already. The trick was that we are not able to set `dataStorage.gppData` to `null`, since the value itself is not nullable (if we don't have anything in the `dataStorage` while trying to access the value it will return an empty map). You can see the solution for that in this PR.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/MessagesApiModel.kt
```diff
@@ -90,20 +90,37 @@ data class Campaigns(
...
-    @SerialName("newUser") val newUser: Boolean?,
-    @SerialName("rejectedAll") val rejectedAll: Boolean?,
-    @SerialName("rejectedCategories") val rejectedCategories: List<String>?,
-    @SerialName("rejectedVendors") val rejectedVendors: List<String>?,
-    @SerialName("signedLspa") val signedLspa: Boolean?,
-    @Serializable(with = CcpaStatusSerializer::class) val status: CcpaStatus?,
-    @Serializable(with = CampaignTypeSerializer::class) override val type: CampaignType,
-    @SerialName("url") override val url: String?,
-    @SerialName("webConsentPayload") val webConsentPayload: JsonObject?,
+    @SerialName("applies")
```
Nevazhnovu, line 93: If I were you, I would put a `@SerialName` tag and value declaration in the same line. It's kinda easier to read for me. But that's 100% optional :) Is it against kotlin code style?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
Nevazhnovu: IMO having `clear` method outside of getter/setter is preferable since it is a separate action and therefore adding it to `get`/`set` action breaks single responsibility principle 🤓

Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
bohdan-go-wombat: @Nevazhnovu that is also a good point, and since in the near future (I assume) we are going to separate `GPP` from the `CPPA`, we will definitely rethink the approach for that

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/MessagesApiModel.kt
bohdan-go-wombat: well, actually from my pov it is better not only to move those to a new line, but also separate one line between parameters, to make it more visible, but I will do that when I'm going to separate those models in the future

Kotlin does not have a strict code style for that, so again, it is only the matter of taste :)

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
andresilveirah: I'd usually agree with separating the two things. However, it's part of the business logic that we _always_ want to clear gpp data and _then_ set its new values.
The problem with separating the two functionalities (clearing + set) is that it makes it easier for anyone to forget clearing it first.

andresilveirah:
@bohdan-go-wombat let's just make `gppData` nullable on CCPADataStorage

andresilveirah (approved):

PR #683, opened 2023-08-22
[DIA-2479] Allow user to set GPP Config

bohdan-go-wombat:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/creation/SpConfigDataBuilder.kt
```diff
@@ -23,6 +24,7 @@ class SpConfigDataBuilder {
     var campaignsEnv: CampaignsEnv = CampaignsEnv.PUBLIC
     var messageTimeout: Long = 5000
     var logger: Logger? = null
+    var gppConfig: GppConfig? = null
 
     operator fun CampaignType.unaryPlus() {
```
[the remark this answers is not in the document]
bohdan-go-wombat: I actually see no advantage in overloading `unaryPlus` in this case. I think the idea was to make it almost intuitive in terms of readability, but it made it less readable from my point of view.

Here is also the response from the ChatGPT:

> The advantage of overloading the unaryPlus function is that it provides you with the ability to define a custom action that should be performed when the + unary operator is used with an instance of your class. This can be particularly useful in scenarios where you want to create more expressive and intuitive APIs for your classes.

And again, from my point of view, it actually doesn't make it _"more expressive and intuitive"_.

But since it works and it is the part of our public API, my proposition is to leave it **as is** for now.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exposed/gpp/GppOptionTernary.kt
```diff
@@ -1,6 +1,6 @@
-package com.sourcepoint.cmplibrary.gpp.dto
+package com.sourcepoint.cmplibrary.exposed.gpp
 
-internal enum class GppTernaryType(
```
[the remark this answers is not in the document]
bohdan-go-wombat: Yes, it will be a part of our public API, since the main goal of the task is to provide our users with the ability to set their own `GppConfig`, with proper `GppOptions`

Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -443,13 +456,17 @@ private class ServiceImpl(
...
+            )
+
             val getChoiceParamReq = GetChoiceParamReq(
                 choiceType = at.toChoiceTypeParam(),
                 accountId = spConfig.accountId.toLong(),
                 propertyId = spConfig.propertyId.toLong(),
                 env = env,
                 metadataArg = campaignManager.metaDataResp?.toChoiceMetaData()?.copy(gdpr = null),
-                includeData = IncludeData.generateIncludeDataForGetChoice(),
+                includeData = getChoiceIncludeData,
```
andresilveirah, line 469: Since we don't do anything with the includeData variables, I'd get back to inlining all these includeData assignments, like:
```kotlin
includeData = IncludeData.generateIncludeDataForGetChoice(
    includeDataGppParam = spConfig.gppConfig?.toIncludeDataGppParam(),
)
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exposed/gpp/GppConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    private constructor(
+        builder: Builder
+    ) : this(
+        coveredTransaction = builder.coveredTransaction ?: GppOptionBinary.NO,
+        optOutOptionMode = builder.optOutOptionMode ?: GppOptionTernary.NOT_APPLICABLE,
+        serviceProviderMode = builder.serviceProviderMode ?: GppOptionTernary.NOT_APPLICABLE,
+    )
+
+    class Builder {
```
andresilveirah, line 20: What's the advantage of having a builder for a data class with only 3 attributes?

andresilveirah (changes requested):
@bohdan-go-wombat I believe we'll need to change the code a bit. 

The goal is to have the `GPPConfig` added to `includeData` by default whenever a client has a CCPA campaign in their config. If no CCPA campaign is added, the config should be null (and not added to `includeData`).

The default `GPPConfig` should be an empty object (all it attributes are null).

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exposed/gpp/GppConfig.kt
bohdan-go-wombat: @andresilveirah I created the `Builder` for the `GppConfig` for several reasons:
1. I know it is a good rule of thumb to use creational patterns in order to allow the users of our API's (in our case - our library) to create objects from our API (I am a user of multiple Android libraries, and whenever the library has such a mechanism it gives me a good UX)
2. It provides an option to use this `Builder`, but it is not mandatory, so our users can choose between 2
3. It provides the user an ability to create `GppConfig` with only a couple of parameters (the same with the constructor, but again UX matter)

To summarize, the main goal was to provide the user with more conventional way to instatiate objects from the external API, without making it **mandatory**.

andresilveirah: I think the added complexity is not worth the benefit. We don't need 2 ways to instantiate the same simple object. The constructor with default values for each attribute is sufficient for ours (and our clients) needs.

The following api is just what we need:
```kotlin
data class GppConfig(
    val a: String? = null,
    val b: String? = null,
    val c: String? = null
)

val config = GppConfig(a: "value")
```

The user can set any value when instantiating the config or leave them untouched.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
andresilveirah:
```kotlin
val getChoiceParamReq = GetChoiceParamReq(
  ...
  includeData = IncludeData.generateIncludeDataForGetChoice(
      includeDataGppParam = spConfig.gppConfig?.toIncludeDataGppParam(),
  )
  ...
)
```

The above looks perfectly fine to me. Let's inline it then.

andresilveirah (approved):

PR #685, opened 2023-08-29
[DIA-2542] Local data versioning

bohdan-go-wombat:
JIRA - https://sourcepoint.atlassian.net/browse/DIA-2542


Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageImpl.kt
```diff
@@ -47,6 +49,14 @@ private class DataStorageImpl(
         PreferenceManager.getDefaultSharedPreferences(context)
     }
 
+    override var localDataVersion: Int
+        get() = preference
+            .getInt(KEY_LOCAL_DATA_VERSION, 1)
```
andresilveirah, line 54: The correct value for when the key is not found should be 0, so we make sure returning users will call `consent-stauts` at least once (and therefore get GPP data).

```kotlin
  .getInt(KEY_LOCAL_DATA_VERSION, 0)
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageImpl.kt
```diff
@@ -169,6 +179,10 @@ private class DataStorageImpl(
         return preference.getString(LOCAL_STATE, null)
     }
 
+    override fun updateLocalDataVersion() {
```
andresilveirah, line 182: I don't see the need to implement this method. 

Whichever class is using `DataStorage` can simply set the value to be the hardcoded one.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -164,6 +164,7 @@ private class ServiceImpl(
 
                 networkClient.getConsentStatus(consentStatusParamReq)
                     .executeOnRight { consentStatusResponse ->
+                        dataStorage.updateLocalDataVersion()
```
andresilveirah, line 167: Here, for example, I'd just set instead of using the dedicated `updatedLocalDataVersion()`

```kotlin
dataStorage.localDataVersion = HARDCODED_LOCAL_DATA_VERSION
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageImpl.kt
bohdan-go-wombat: makes sense, wasn't sure if we want to enforce users to get GPP data right after we release it

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
bohdan-go-wombat: we have to encapsulate the logic of updating `localDataVersion` to the `HARDCODED_LOCAL_DATA_VERSION`

andresilveirah (approved):

andresilveirah (approved):

PR #688, opened 2023-09-05
[DIA-2628] Needs to handle no-internet better

bohdan-go-wombat:
### Task
Jira - https://sourcepoint.atlassian.net/browse/DIA-2628

### Description
This PR introduces additional checks that handles if the device has/lost connection to the Internet before critical BE calls, and handle those in a manner of throwing particular exception. Also, this PR introduces additional fix for manifest files, which fixes an issue with Android 12+.


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -134,6 +136,16 @@ internal class SpConsentLibImpl(
     }
 
     private fun localLoadMessage(authId: String?, pubData: JSONObject?, cmpViewId: Int?) {
+
+        if (context.isInternetConnected().not()) {
+            spClient.onError(
```
andresilveirah, line 141: @bohdan-go-wombat what do you think about moving these checks to `spClient`? This way we could check it on before _every_ call?

bohdan-go-wombat: if we want to do so, we should move it either to the `NetworkClient` or to the `chech { ... }` method

andresilveirah: oh, my bad, I meant to say `NetworkClient`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/extensions/ContextExt.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.content.Context
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.os.Build
+
+/**
+ * Extension function that checks whether or not the device is connected to the Internet
+ */
+fun Context.isInternetConnected(): Boolean {
```
Nevazhnovu, line 11: @bohdan-go-wombat just curious, have you tested the `fun` performance while throttling network or using vpn?

Nevazhnovu: like for this extension by the way  👍

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
```diff
@@ -12,6 +12,7 @@ import android.webkit.JavascriptInterface
 import android.webkit.WebChromeClient
 import android.webkit.WebSettings
 import android.webkit.WebView
+import com.example.cmplibrary.R
```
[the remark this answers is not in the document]
bohdan-go-wombat: a class that contains resources, in our case we need string resource with the description of the network error

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/extensions/ContextExt.kt
bohdan-go-wombat: @Nevazhnovu I haven't tested using VPN or by throttling my network, but I tested on WiFi and by using celular connection

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
bohdan-go-wombat: @andresilveirah I think that makes sense, I was just not really confident about passing context instance that much, because that might cause memory leaks

but let me try that out

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
```diff
@@ -107,8 +109,8 @@ internal class ConsentWebView(
         consent: String?,
     ): Either<Boolean> = check {
 
-        if (connectionManager.isConnected.not()) throw NoInternetConnectionException(
-            description = "No internet connection"
+        if (context.isInternetConnected().not()) throw NoInternetConnectionException(
+            description = context.getString(R.string.exception_no_internet_connection_text)
```
andresilveirah, line 113: @bohdan-go-wombat could you move the description to be a static (or regular) field of `NoInternetConnectionException` ? This way we avoid having a dependency on `R` in our code. Remember this is likely to get replaced by the core module and the core module probably won't have access to R.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
```diff
@@ -190,6 +196,18 @@ internal class ConsentWebView(
...
+                jsClientLib.onError(
+                    view = this@ConsentWebView,
+                    error = NoInternetConnectionException(
+                        description = context.getString(R.string.exception_no_internet_connection_text)
+                    )
+                )
+                jsClientLib.dismiss(this@ConsentWebView)
+                return
+            }
+
```
andresilveirah, line 210: @bohdan-go-wombat if the network requests are checking for connectivity, I believe we don't need to check on `onAction`.

bohdan-go-wombat: @andresilveirah this is the case when the user had connection when getting a First Layer Message, but lost it before taking an action, so for the sake of additional sanity check we should leave it as is

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
bohdan-go-wombat: @andresilveirah I guess you are correct on this one, since we are not going to translate this description into any other language

although, it is a good practise to use string constants from the string resources rather than hardcoding it in a bunch of places (hardcoding makes it harder to change the value of this string in the future)

andresilveirah (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
bohdan-go-wombat: @andresilveirah deleted

PR #689, opened 2023-09-08
[DIA-2066] `OnBackPressed()` for ConsentWebview 

Nevazhnovu:
This works for `pre-prod` - only for now

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
@@ -2,7 +2,6 @@ package com.sourcepoint.cmplibrary
 
 import android.content.Context
 import android.view.View
-import android.webkit.WebView
 import com.sourcepoint.cmplibrary.campaign.CampaignManager
 import com.sourcepoint.cmplibrary.consent.ClientEventManager
 import com.sourcepoint.cmplibrary.consent.ConsentManager
@@ -419,6 +418,7 @@ internal class SpConsentLibImpl(
     }
 
     override fun dispose() {
+        webview = null
         executor.dispose()
         viewManager.removeAllViews()
     }
@@ -432,6 +432,8 @@ internal class SpConsentLibImpl(
             """.trimIndent(), null)
         }
     }
+
+    override fun isWebviewShown(): Boolean = (webview as ConsentWebView).isShown
 
     /** Start Receiver methods */
     inner class JSReceiverDelegate : JSClientLib {
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
@@ -171,7 +171,7 @@ internal class SpConsentLibImpl(
                             )
                                 .executeOnLeft { spClient.onError(it) }
                                 .getOrNull()
-                            this.webview = webView
+//                            this.webview = webView
 
                             /** inject the message into the WebView */
                             val url = firstCampaign2Process.url
@@ -427,9 +427,12 @@ internal class SpConsentLibImpl(
 
     override fun onBackPressed() {
         webview.let {
-            (webview as ConsentWebView).evaluateJavascript("""
+            (webview as ConsentWebView).evaluateJavascript(
+                """
                 window.postMessage({ name: 'sp.BACK' })
-            """.trimIndent(), null)
+                """.trimIndent(),
+                null
+            )
         }
     }
 
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
@@ -171,7 +171,7 @@ internal class SpConsentLibImpl(
                             )
                                 .executeOnLeft { spClient.onError(it) }
                                 .getOrNull()
-//                            this.webview = webView
+                            this.webview = webView
 
                             /** inject the message into the WebView */
                             val url = firstCampaign2Process.url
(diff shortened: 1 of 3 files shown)


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -415,12 +418,26 @@ internal class SpConsentLibImpl(
...
+            (webview as ConsentWebView).evaluateJavascript(
+                """
+                window.postMessage({ name: 'sp.BACK' })
+                """.trimIndent(),
+                null
+            )
+        }
+    }
+
+    override fun isWebviewShown(): Boolean = (webview as ConsentWebView).isShown
```
bohdan-go-wombat, line 439: i think here we can add an extra safety check, something like `(webview as? ConsentWebView)?.isShown ?: false`, since `webview` here might be nullable, so extra safety is never an overkill

bohdan-go-wombat (approved):

samples/metaapp/src/main/java/com/sourcepointmeta/metaapp/tv/demo/DemoActivityTv.kt
```diff
@@ -83,6 +83,13 @@ class DemoActivityTv : FragmentActivity() {
         }
     }
 
+    override fun onBackPressed() {
+        if (spConsentLib.isWebviewShown().not())
+            super.onBackPressed()
+        else
+            spConsentLib.onBackPressed()
+    }
```
andresilveirah, line 91: @Nevazhnovu does that mean the application has to implement that? Since we own the view and we have access to the activity itself, can't the SDK override the `onBackPressed` and have it transparent to the host app?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -58,6 +59,7 @@ internal class SpConsentLibImpl(
 
     private val remainingCampaigns: Queue<CampaignModel> = LinkedList()
     private var currentNativeMessageCampaign: CampaignModel? = null
+    private var webview: IConsentWebView? = null
```
andresilveirah, line 62: @Nevazhnovu have a look at the `ViewManager.kt` file. The `ViewsManagerImpl` class has a reference to the SDK's main view 
```kotlin
    val mainView: ViewGroup?
        get() = weakReference.get()?.findViewById(R.id.content)
``` 

I recommend moving the logic there and using that reference, instead a new one in `SpConsentLib`.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -415,12 +418,26 @@ internal class SpConsentLibImpl(
...
+            (webview as ConsentWebView).evaluateJavascript(
+                """
+                window.postMessage({ name: 'sp.BACK' })
+                """.trimIndent(),
+                null
+            )
+        }
+    }
+
+    override fun isWebviewShown(): Boolean = (webview as? ConsentWebView)?.isShown ?: false
```
andresilveirah, line 439: Also, is there a way to implement this logic without checking if the webview is shown?

andresilveirah (changes requested):

samples/metaapp/src/main/java/com/sourcepointmeta/metaapp/tv/demo/DemoActivityTv.kt
bohdan-go-wombat: @andresilveirah but we also don't want to prohibit the user from being able to utilize their application's `onBackPressed()`

meaning that:
• WHEN there are no `webView` on the screen
• AND the user clicks back button (maybe to go out from the screen were our `webView` was triggered)
• THEN the user should be redirected from that screen

and also we should be careful about this case:
• WHEN there is a visible `webView` on the screen
• AND the user clicks back button and expects to go back inside the `webView`
• THEN the `onBackPressed()` should _ONLY_ be triggered for the `webView` and not for the entire activity

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
bohdan-go-wombat: @andresilveirah not sure about that, since we have to prohibit activity's `onBackPressed()` when the `webView` is shown and vice versa, so we need a flag that indicates whether the `webView` is shown or not

samples/metaapp/src/main/java/com/sourcepointmeta/metaapp/tv/demo/DemoActivityTv.kt
Nevazhnovu: @andresilveirah AFAIK in Android only Activity/Context which is `DemoActivityTv` can overload the `onBackPressed()` method.  @bohdan-go-wombat Proof me wrong if I am

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
Nevazhnovu: @andresilveirah It would totally make sense if we could embed `onBackPressed()` and make it internal. But currently, we still need to make that public in `SpConsentLibImpl.kt` in order for the user can access that.

carmelo-iriti:
This fix we want to create doesn't seem right to me, it looks we are trying to fix from an example app a problem which is probably in the rendering app, @Nevazhnovu did you ask @MarkGeeRomano about it?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
carmelo-iriti: Calling the onBackPressed from SDK means that we have a reference back to the host, our SDK is a simple library which is just providing 2 APIs, regardless of the host component and we don't want to interfere with the lifecycle of the host object (in our case the Activity or, in a different scenario, a simple view).
This fix won't work if the host object is a fragment or a tv view component

carmelo-iriti:
The back mechanism should be implemented in easier and smoother way, I'm going to close this pr since the base branch is totally different

PR #693, opened 2023-09-21
[DIA-2646] Fix for GDPR applies not being updated

bohdan-go-wombat:
Testing wise, here are the response of the `/meta-data` request with the actual value that the user receives in `onSpFinished()` callback:

image


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModelExt.kt
```diff
@@ -15,12 +15,12 @@ import kotlinx.serialization.json.jsonPrimitive
 internal fun GdprCS.toGDPRUserConsent(): GDPRConsentInternal {
     return GDPRConsentInternal(
         uuid = uuid,
+        applies = applies ?: false,
         tcData = JsonObject(TCData ?: emptyMap()),
         grants = grants ?: emptyMap(),
         euconsent = euconsent ?: "",
         acceptedCategories = grants?.toAcceptedCategories()?.toList(),
         childPmId = null,
-        applies = TCData?.fromTcDataToGdprApplies(),
```
andresilveirah, line 23: Nice catch.

andresilveirah (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -490,6 +490,11 @@ private class CampaignManagerImpl(
 
         // handle gdpr
         response.gdpr?.apply {
+            gdprConsentStatus?.let { gdprCS ->
+                val updatedGdprConsentStatus = gdprCS.copy(applies = applies)
```
Nevazhnovu, line 494: @bohdan-go-wombat I suggest renaming`CS` to `ConsentStatus`

Nevazhnovu (approved):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
bohdan-go-wombat: @Nevazhnovu in that case we will have shadowed naming of `gdprConsentStatus` from the `CampaignManager` itself, which acceptable, but not really good from the readability point of view, since you have 2 variables with the same name

although, they both mean the same thing (only that the one inside curly brackets will just be non nullable), I just think that `CS` abbreviation is pretty obvious in this case

PR #694, opened 2023-09-25
[DIA-2756] Runtime Args update with the new version of the gradle plugin

carmelo-iriti:


carmelo-iriti:
it was enough to switch locally the java version

PR #695, opened 2023-09-26
[DIA-2762] UI tests fix

carmelo-iriti:


samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinNativeMessage.kt
```diff
@@ -37,32 +32,11 @@ import org.koin.core.context.loadKoinModules
 class MainActivityNativeMessTest {
 
     lateinit var scenario: ActivityScenario<NativeMessageActivity>
-
-    @MockK
-    lateinit var spClient: SpClient
```
bohdan-go-wombat, line 42: just a general question, because I know that this PR is basically a big _REVERT_ to the old state:
is there any advantage in initializing mocks in each and every test case, instead of initializing it like the way we do right now?

bohdan-go-wombat (approved):
revert that returns green checkmarks back, looks good to me 🤞

samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinNativeMessage.kt
ciriti: Yes, mock objects retain their state, this means that the checks after the first usage give you the wrong outcome

PR #696, opened 2023-10-02
[DIA-2628] Needs to handle no-internet better

bohdan-go-wombat:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/NetworkClientImpl.kt
```diff
@@ -30,13 +33,15 @@ import okhttp3.Request
 import okhttp3.RequestBody
 
 internal fun createNetworkClient(
+    context: Context,
```
carmelo-iriti, line 36: This logic should not be in the NetworkClient, it should be in the Service class (ServiceImpl).

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exception/ExceptionDescriptions.kt
```diff
@@ -0,0 +1,8 @@
+package com.sourcepoint.cmplibrary.exception
+
+class ExceptionDescriptions {
+
+    companion object {
```
carmelo-iriti, line 5: We already have an object for this purpose ExceptionCodes.kt

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/creation/CreationalUtility.kt
```diff
@@ -34,13 +35,17 @@ internal fun createLogger(errorMessageManager: ErrorMessageManager): Logger {
     )
 }
 
-internal fun networkClient(netClient: OkHttpClient, responseManage: ResponseManager, logger: Logger): NetworkClient {
-    return createNetworkClient(
+internal fun networkClient(
```
carmelo-iriti, line 38: no need this change

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
```diff
@@ -107,9 +106,7 @@ internal class ConsentWebView(
         consent: String?,
     ): Either<Boolean> = check {
 
-        if (connectionManager.isConnected.not()) throw NoInternetConnectionException(
```
carmelo-iriti, line 110: We need the ConnectionManager, the `connectionManager.isConnected` was good

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/creation/Builder.kt
```diff
@@ -78,10 +76,9 @@ class Builder {
         val errorManager = errorMessageManager(campaignManager, client)
         val logger = spc.logger ?: createLogger(errorManager)
         val jsonConverter = JsonConverter.create()
-        val connManager = ConnectionManager.create(appCtx)
```
carmelo-iriti, line 81: you can inject the ConnectionManager into the ServiceImpl class

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/ViewsManager.kt
```diff
@@ -132,7 +128,6 @@ private class ViewsManagerImpl(
             check {
                 ConsentWebView(
                     context = it,
-                    connectionManager = connectionManager,
```
carmelo-iriti, line 135: no need this change

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/ViewsManager.kt
```diff
@@ -155,7 +150,6 @@ private class ViewsManagerImpl(
...
```
carmelo-iriti, line 158: no need this change

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/util/ConnectionExt.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.content.Context
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.os.Build
+
+/**
+ * Extension function that checks whether or not the device is connected to the Internet
+ */
+fun Context.isInternetConnected(): Boolean {
```
carmelo-iriti, line 11: This is good!
This should be internal and used from the ConsentManager

carmelo-iriti (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/ViewsManager.kt
bohdan-go-wombat: this change is needed, if we delete `ConnectionManager`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
bohdan-go-wombat: even though it was ok to use `connectionManager.isConnected` previously, the `ConnectionManager` itself did not took into account one deprecated feature (`NetworkInfo` class is deprecated since `API 29`)

when I started adding a check for this deprecated feature, I thought that we do not need a separate object to check Internet connection, since:
1. the only purpose of the `ConnectionManager` is to check Internet connection, which is ok, but nothing more, no checks for a specific connection type or whatever, so there is no need in creating a specific class for that
2. the `ConnectionManager` object uses context instance, for which we can actually create an extension function and that would do the trick

overall, I thought that creating `ConnectionManager` is a bit of overengineering, so that is why I decided to entirely remove `ConnectionManager` and use `Context.isInternetConnected()` extension function

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/creation/CreationalUtility.kt
bohdan-go-wombat: this is just a small refactor to fit to the Kotlin Code Styles, for example:
1. [Class headers﻿](https://kotlinlang.org/docs/coding-conventions.html#class-headers)
2. [Functions](https://kotlinlang.org/docs/coding-conventions.html#functions)
3. etc

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/util/ConnectionExt.kt
bohdan-go-wombat: I have 2 questions here:
1. is `ConsentManager` a part of data layer or not?
2. what is the purpose of using this check specifically in `ConsentManager`?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exception/ExceptionDescriptions.kt
carmelo-iriti: I see your point, in this case just add the "No Internet connection." into the exception since this is the only place we use it

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/exception/Exceptions.kt
```diff
@@ -171,7 +171,7 @@ internal class RenderingAppConnectionTimeoutException @JvmOverloads constructor(
  */
 internal class NoInternetConnectionException @JvmOverloads constructor(
     cause: Throwable? = null,
-    description: String,
+    description: String = ExceptionDescriptions.EXCEPTION_NO_INTERNET_CONNECTION_DESCRIPTION,
```
carmelo-iriti, line 174: just replace it with "No Internet connection."

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/ViewsManager.kt
carmelo-iriti: We cannot delete the ConnectionManager, it is the only way to inject a mock and create UI tests for this feature

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/ViewsManager.kt
carmelo-iriti: the same here

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/core/web/ConsentWebView.kt
carmelo-iriti: The ConnectionManager follows  the Single responsibility principle which we use to isolate the Context. 
The context cannot be replace with a mock object.
Each feature that we add should have a UI test and this is the easy and controllable way of doing it

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/creation/CreationalUtility.kt
carmelo-iriti: Sounds good but in a pr dedicated to a refactor, let’s keep the changes to the minimum in this pr

carmelo-iriti (changes requested):

carmelo-iriti (approved):

carmelo-iriti:
The pr now it's good, before merging we need to check that the graceful degradation works and the sequence of the callbacks is correct

carmelo-iriti:
The tests in the [MainActivityKotlinOldConsentTest.kt](https://github.com/SourcePointUSA/android-cmp-app/pull/696/files#diff-eccec64516e72e96c64920a241d3e3201c9fbf46bd6584518f87f7a8f6270eb8) are failing because a regression which I'm fixing in the DIA-2819 mr

carmelo-iriti (approved):

PR #699, opened 2023-10-04
[DIA-2819] Fix of the migration from v6 to v7

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/MessagesApiModelExt.kt
```diff
@@ -2,11 +2,52 @@ package com.sourcepoint.cmplibrary.data.network.model.optimized
...
+                put("type", "RecordString")
+            }
+            putJsonObject("campaigns") {
+                put("type", "RecordString")
+            }
+            putJsonObject("webConsentPayload") {
+                put("type", "RecordString")
+            }
+            // TODO
+            put("", JsonConverter.converter.encodeToJsonElement(IncludeDataGppParam()))
```
carmelo-iriti, line 38: @bohdan-go-wombat This is the point that needs refactored, we need to re-add the gpp include data.
For now just reconstruct the json, in the next step we can refactor it

bohdan-go-wombat (approved):
revert LGTM 👍

PR #713, opened 2023-10-10
[DIA-2851] Automation test for  the no internet exception

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ConsentManager.kt
```diff
@@ -84,8 +84,10 @@ private class ConsentManagerImpl(
     override var sPConsentsError: ((Throwable) -> Unit)? = null
 
     override val storedConsent: Boolean
```
bohdan-go-wombat, line 86: I believe that the names of the boolean flags should start by `is...`, `has...`, etc

so in this case it is better to name it `hasStoredConsent`, which will indicate whether or not the `DataStorage` contains either gdpr or ccpa consents

bohdan-go-wombat: but this is just a suggestion for now

samples/app/src/main/java/com/sourcepoint/app/v6/di/AppModule.kt
```diff
@@ -39,7 +39,8 @@ val appModule = module {
             customVendorList = customVendorDataList,
             customCategories = customCategoriesData,
             useGdprGroupPmIfAvailable = false,
-            authId = null //get(qualifier = named("authId")))
+            authId = null, //get(qualifier = named("authId")))
+            diagnostic = emptyList()
```
bohdan-go-wombat, line 43: can we leave comma at the end of the constructor param list, just so next time we want to add a new param to the constructor, it will just appear as an addition of one line, and not changing one and adding one?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ConsentManager.kt
carmelo-iriti: Makes sense

bohdan-go-wombat (approved):

PR #714, opened 2023-10-18
[DIA-2757] UI test for the applies condition

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -474,15 +464,11 @@ private class CampaignManagerImpl(
...
-            // regenerate and update US privacy string with new values in the data storage
-            dataStorage.uspstring = generateCcpaUspString(
-                applies = value?.applies,
-                ccpaStatus = value?.status,
-                signedLspa = value?.signedLspa,
-            )
+            dataStorage.run {
+                ccpaConsentStatus = serialised
+                gppData = value?.gppData
+                uspstring = value?.uspstring
```
bohdan-go-wombat, line 470: I'm not sure if we need this one, since we don't want to count on the `uspString` from any backend response and we are going to generate `uspString` anyways, but maybe I'm missing something here

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -540,8 +526,10 @@ private class CampaignManagerImpl(
         response.ccpa?.apply {
             applies?.let { i ->
                 ccpaConsentStatus?.let { ccpaCS ->
-                    val updatedCcpaConsentStatus = ccpaCS.copy(applies = i)
-                    ccpaConsentStatus = updatedCcpaConsentStatus
+                    val updatedCcpaCS = ccpaCS.copy(applies = i)
+                    // update the new uspstring value based on the applies
+                    val uspstring = updateCcpaUspString(updatedCcpaCS, logger)
+                    ccpaConsentStatus = updatedCcpaCS.copy(uspstring = uspstring)
```
bohdan-go-wombat, line 532: like for this one 👍

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: We are going to update the `uspString` just once and only in one place, which is in the the method to handle the logic for meta data, this is just a simple assignment, it is just a set

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -50,93 +49,46 @@ data class ConsentStatusResp(
...
+    @SerialName("rejectedAll") val rejectedAll: Boolean?,
+    @SerialName("rejectedCategories") val rejectedCategories: List<String>?,
+    @SerialName("rejectedVendors") val rejectedVendors: List<String>?,
+    @SerialName("signedLspa") val signedLspa: Boolean?,
+    @SerialName("uspstring") val uspstring: String? = null,
+    @Serializable(with = CcpaStatusSerializer::class) val status: CcpaStatus?,
+    @SerialName("GPPData") @Serializable(with = JsonMapSerializer::class) val gppData: Map<String, JsonElement>? = null,
+    @SerialName("uuid") var uuid: String?,
+    @SerialName("webConsentPayload") val webConsentPayload: JsonObject? = null,
+)
```
bohdan-go-wombat, line 66: from my pov, the thing that you refactored here looked more readable and it followed [KCS](https://kotlinlang.org/docs/coding-conventions.html#annotations), but it might be the matter of taste here, so I'm ok with that

cmplibrary/src/preprod/java/com/sourcepoint/cmplibrary/creation/ComponentFactory.kt
```diff
@@ -1,7 +1,22 @@
...
 import com.sourcepoint.cmplibrary.data.network.connection.create
+import com.sourcepoint.cmplibrary.data.network.createNetworkClient
+import com.sourcepoint.cmplibrary.data.network.util.HttpUrlManagerSingleton
+import com.sourcepoint.cmplibrary.data.network.util.ResponseManager
+import com.sourcepoint.cmplibrary.exception.Logger
+import okhttp3.OkHttpClient
 
 internal fun getConnectionManager(appCtx: Context): ConnectionManager = ConnectionManager.create(appCtx)
+
+internal fun networkClient(appCtx: Context, netClient: OkHttpClient, responseManage: ResponseManager, logger: Logger): NetworkClient {
```
bohdan-go-wombat, line 15: why do we need and instance of `Context` here, since we don't use it?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/SpUtils.kt
```diff
@@ -24,11 +25,13 @@ import kotlinx.serialization.decodeFromString
 
 private const val DEFAULT_CCPA_USP_STRING = "1---"
 
-fun generateCcpaUspString(
-    applies: Boolean?,
-    ccpaStatus: CcpaStatus?,
-    signedLspa: Boolean?,
+internal fun updateCcpaUspString(
```
bohdan-go-wombat, line 28: big thumbs up for this refactor 👍

cmplibrary/src/preprod/java/com/sourcepoint/cmplibrary/creation/ComponentFactory.kt
carmelo-iriti: because using the variants technique, we must have the signature of the method the same in all files (the contents can be different)

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModelExt.kt
```diff
@@ -31,6 +31,7 @@ internal fun CcpaCS.toCCPAConsentInternal(): CCPAConsentInternal {
         thisContent = JSONObject(),
         signedLspa = signedLspa,
         webConsentPayload = webConsentPayload,
+        uspstring = this.uspstring ?: "1YNN"
```
bohdan-go-wombat, line 34: can we use constant here, instead of hardcoded value?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/MessagesApiModelExt.kt
```diff
@@ -131,26 +131,25 @@ internal fun MessagesParamReq.toConsentStatusParamReq(
     )
 }
 
-internal fun CCPA.toCcpaCS() = CcpaCS(
+internal fun CCPA.toCcpaCS(applies: Boolean?) = CcpaCS(
```
bohdan-go-wombat, line 134: now I'm thinking, wouldn't it be better for us not to update `applies` value anywhere, except from the response of the `/meta-data` call? that will ease-out this flow, so that we don't need to pass an extra value into the ext function, and we will not worry about the value of `applies` having wrong value

wdyt?

cmplibrary/src/release/java/com/sourcepoint/cmplibrary/creation/ComponentFactory.kt
```diff
@@ -1,7 +1,22 @@
...
 import com.sourcepoint.cmplibrary.data.network.connection.create
+import com.sourcepoint.cmplibrary.data.network.createNetworkClient
+import com.sourcepoint.cmplibrary.data.network.util.HttpUrlManagerSingleton
+import com.sourcepoint.cmplibrary.data.network.util.ResponseManager
+import com.sourcepoint.cmplibrary.exception.Logger
+import okhttp3.OkHttpClient
 
 internal fun getConnectionManager(appCtx: Context): ConnectionManager = ConnectionManager.create(appCtx)
+
+internal fun networkClient(appCtx: Context, netClient: OkHttpClient, responseManage: ResponseManager, logger: Logger): NetworkClient {
```
bohdan-go-wombat, line 15: same here

bohdan-go-wombat (commented):
I have a couple of questions, after we sort those out I can approve

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
andresilveirah: Not sure I understand the full context. But uspString needs to be updated in more places than only in the response of `/meta-data`. 

What happens if the user opens the CCPA pm and rejects all, for example?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
andresilveirah: I'd go ahead and remove the annotations all together, keeping only the ones that are strictly necessary (those whose attribute name differ from the api response). 

@bohdan-go-wombat thanks for sharing the KCS link, it seems that annotations without arguments are allowed to be in the same line as the example:
```kotlin
@Test fun foo() { /*...*/ }
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModelExt.kt
carmelo-iriti: makes sense

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/SpUtils.kt
```diff
@@ -37,8 +40,27 @@ fun generateCcpaUspString(
             else -> "N"
         }
         val lspaCoveredTransaction = if (signedLspa == null || signedLspa == false) "N" else "Y"
-        "$specificationVersion$opportunityToOptOut$optOutSale$lspaCoveredTransaction"
-    } else DEFAULT_CCPA_USP_STRING
+        val usPString = "$specificationVersion$opportunityToOptOut$optOutSale$lspaCoveredTransaction"
+        logger?.computation(
```
andresilveirah, line 44: @carmelo-iriti these loggers are really getting on the way of readability. Can we get rid of it?

cmplibrary/src/preprod/java/com/sourcepoint/cmplibrary/creation/ComponentFactory.kt
bohdan-go-wombat: gotcha

andresilveirah (changes requested):

andresilveirah:
@carmelo-iriti I believe this PR should have been broken down. The title says UI tests for applies condition, but I see changes for the usp string, logic changes, tests and code styling changes all in the same group.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/SpUtils.kt
carmelo-iriti: Unfortunately no, because I use them to check if the conditions are matched and in production they are not logging anything (using the veriants technique)

carmelo-iriti:
@andresilveirah the changes for the uspstring have to stay in order to make the tests working

andresilveirah:
> @andresilveirah this is the original code styling which was changed in some pr

Agree. Many things were changed in the previous PRs. And they were changed for a reason. I suggest we keep things the way they are unless there's a strong argument against it.

carmelo-iriti:
@andresilveirah  originally I chose the standard code style for kotlin where the annotations are in line with the variable different than java where every time you have a new line

andresilveirah:
@carmelo-iriti understood, thanks for the info. Do you mind reverting back the styling changes just so we can focus on the actual changes in this PR? If necessary, we can address code styling changes in a separate PR only for that.

carmelo-iriti:
@andresilveirah ok

cmaurersp:
The one issue I saw with the applies value is when you click on the Save & Close button in the privacy manager the onConsentReady and onSpFinish. This issue was submitted on https://sourcepoint.atlassian.net/browse/DIA-2901

cmaurersp:
The consent.ccpa?.consents?.uspstring and the consent.ccpa?.consents?.applies were aligned in my tests

carmelo-iriti:
@cmaurersp Thanks, I skipped the UI test for the gdpr PM 👍🏻

PR #724, opened 2023-10-25
[DIA-2918] Consentstatus missing

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/exposed/SPConsents.kt
```diff
@@ -51,6 +53,7 @@ internal data class GDPRConsentInternal(
     override var grants: Map<String, GDPRPurposeGrants> = emptyMap(),
     override val acceptedCategories: List<String>? = null,
     override val applies: Boolean = false,
+    override val consentStatus: ConsentStatus? = null,
```
bohdan-go-wombat, line 56: I took a look at the previous implementation, and it seems like there was an empty object as a default value of `consentStatus`, like here:
```
override val consentStatus: ConsentStatus = ConsentStatus(),
```
not sure if that was a business requirement, but I think that is something worth mentioning

carmelo-iriti: I see, adding a default value can be misleading because it can add false information to the consent status

bohdan-go-wombat (approved):

PR #729, opened 2023-10-26
[DIA-2886] Fix Gdpr applies

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageImpl.kt
```diff
@@ -99,8 +99,15 @@ private class DataStorageImpl(
      * By default - false (if there are no ccpaConsentStatus in data storage or applies value is null)
      */
     override val ccpaApplies: Boolean
-        get() = ccpaConsentStatus?.let { verifiedCcpaConsentStatusString ->
-            check { JsonConverter.converter.decodeFromString<CcpaCS>(verifiedCcpaConsentStatusString) }
+        get() = metaDataResp?.let { metaData ->
```
bohdan-go-wombat, line 102: I like that, since the most relevant info about the campaigns applies value is indeed in the `/meta-data` response, it makes perfect sense to depend on that

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdpr.kt
```diff
@@ -335,11 +321,3 @@ private class DataStorageGdprImpl(context: Context) : DataStorageGdpr {
 
     private fun fail(param: String): Nothing = throw RuntimeException("$param not fund in local storage.")
 }
-
-internal fun DataStorageGdpr.getGDPRConsent(): Either<GDPRConsentInternal> = check {
```
bohdan-go-wombat, line 339: was it unused?

bohdan-go-wombat (approved):

PR #731, opened 2023-10-31
[DIA-2896] Implement consent expiration

carmelo-iriti:


cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/campaign/CampaignManagerImplTest.kt
```diff
@@ -178,4 +181,20 @@ class CampaignManagerImplTest {
 
         cm.metaDataResp.assertNull()
     }
+
+    @Test
+    fun `GIVEN_an_expired_GDPR_and_a_valid_CCPA_campaign_DELETE_only_the_GDPR_consent`() {
+        val json = JSONObject("v7/expired_gdpr_valid_ccpa.json".file2String())
+        appContext.storeTestDataObj(json.toList())
+        ds.deleteGdprConsent()
+        appContext.spEntries().toList().find { it.first.contains("GDPR") }.assertNull()
```
bohdan-go-wombat, line 190: I think we should assert that `CCPA` campaign is `not null` in this case, since we want to make sure that only `GDPR` campaign was deleted

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/campaign/CampaignManagerImplTest.kt
```diff
@@ -178,4 +181,20 @@ class CampaignManagerImplTest {
...
+    }
+
+    @Test
+    fun `GIVEN_an_expired_CCPA_and_a_valid_GDPR_campaign_DELETE_only_the_GDPR_consent`() {
+        val json = JSONObject("v7/expired_ccpa_valid_gdpr.json".file2String())
+        appContext.storeTestDataObj(json.toList())
+        ds.deleteCcpaConsent()
+        appContext.spEntries().toList().find { it.first.contains("CCPA") }.assertNull()
```
bohdan-go-wombat, line 198: I think we should assert that `GDPR` campaign is `not null` in this case, since we want to make sure that only `CCPA` campaign was deleted

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpaImplTest.kt
```diff
@@ -17,21 +17,21 @@ class DataStorageCcpaImplTest {
     fun check_DataStorage_TcData_gets_stored() {
         // Context of the app under test.
         val appContext = InstrumentationRegistry.getInstrumentation().targetContext
-        val storage = DataStorageCcpa.create(appContext).apply { clearAll() }
+        val storage = DataStorageCcpa.create(appContext).apply { deleteCcpaConsent() }
         val map = TreeMap<String, String>()
     }
```
bohdan-go-wombat, line 22: I don't get what this test case is actually testing, it seems like we initializing `dataStorage`, deleting CCPA consent on a go, and then do nothing

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpaImplTest.kt
```diff
@@ -17,21 +17,21 @@ class DataStorageCcpaImplTest {
...
 
     @Test
     fun clear_data_DataStorage() {
```
bohdan-go-wombat, line 25: I think we should rename the test case name to something more descriptive of what the test is actually doing, something like
> given_delete_ccpa_consent_called_with_stored_consent_then_should_clear_stored_consent

or something along those lines

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdprImplTest.kt
```diff
@@ -13,7 +13,7 @@ class DataStorageGdprImplTest {
     fun clear_data_DataStorage() {
```
bohdan-go-wombat, line 13: I think the same should be here, just for the sake of getting more descriptive in the namings of our test cases

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -650,4 +652,45 @@ private class CampaignManagerImpl(
             ccpaCS = ccpaConsentStatus
         )
     }
+
+    override val isGdprExpired: Boolean
+        get() {
+
+            val formatter = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.getDefault())
+            val gdprExpirationDate = dataStorage.gdprExpirationDate?.let { formatter.parse(it) } ?: return false
```
bohdan-go-wombat, line 660: I think `SimpleDateFormat.parse()` might throw `ParseException`, we should keep that in mind, maybe even wrap up this call in `try/catch` and return `false` if the exception being thrown

BUT, that might give us a false positive result, when we will not be able to get the issue...

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -650,4 +652,45 @@ private class CampaignManagerImpl(
...
+            )
+
+            return isGdprExpired
+        }
+
+    override val isCcpaExpired: Boolean
+        get() {
+
+            val formatter = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.getDefault())
+            val ccpaExpirationDate = dataStorage.ccpaExpirationDate?.let { formatter.parse(it) } ?: return false
```
bohdan-go-wombat, line 678: same here

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpa.kt
```diff
@@ -267,7 +283,7 @@ private class DataStorageCcpaImpl(context: Context) : DataStorageCcpa {
                 .apply()
         }
 
-    override fun clearAll() {
+    override fun deleteCcpaConsent() {
```
bohdan-go-wombat, line 286: kudos for the renaming of this 👍

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpa.kt
```diff
@@ -70,6 +74,7 @@ internal interface DataStorageCcpa {
         const val CCPA_DATE_CREATED = "sp.ccpa.key.date.created"
         const val CCPA_SAMPLING_VALUE = "sp.ccpa.key.sampling"
         const val CCPA_SAMPLING_RESULT = "sp.ccpa.key.sampling.result"
+        const val CCPA_EXPIRATION_DATE_MESSAGE = "sp.ccpa.key.expiration.date"
```
bohdan-go-wombat, line 77: can we name figure out some sort of a convention for the naming of the constants, for example if we have `sp.ccpa.key.expiration.date`, then the naming will be something like `KEY_CCPA_EXPIRATION_DATE`?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdpr.kt
```diff
@@ -97,6 +101,7 @@ internal interface DataStorageGdpr {
         const val GDPR_DATE_CREATED = "sp.gdpr.key.date.created"
         const val GDPR_SAMPLING_VALUE = "sp.gdpr.key.sampling"
         const val GDPR_SAMPLING_RESULT = "sp.gdpr.key.sampling.result"
+        const val GDPR_EXPIRATION_DATE_MESSAGE = "sp.gdpr.key.expiration.date"
```
bohdan-go-wombat, line 104: same here, I believe it would be better if we stick to some sort of a convention for the namings of the constants, like `KEY_GDPR_EXPIRATION_DATE`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdpr.kt
```diff
@@ -266,7 +282,7 @@ private class DataStorageGdprImpl(context: Context) : DataStorageGdpr {
                 .apply()
         }
 
-    override fun clearAll() {
+    override fun deleteGdprConsent() {
```
bohdan-go-wombat, line 285: 👍

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -62,7 +62,8 @@ data class CcpaCS(
     @Serializable(with = CcpaStatusSerializer::class) val status: CcpaStatus?,
     @SerialName("GPPData") @Serializable(with = JsonMapSerializer::class) val gppData: Map<String, JsonElement>? = null,
     @SerialName("uuid") var uuid: String?,
-    @SerialName("webConsentPayload") val webConsentPayload: JsonObject? = null
+    @SerialName("webConsentPayload") val webConsentPayload: JsonObject? = null,
+    @SerialName("expirationDate") var expirationDate: String?,
```
bohdan-go-wombat, line 66: thanks for the comma at the end 👍

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -89,6 +90,7 @@ data class GdprCS(
     @SerialName("uuid") var uuid: String?,
     @SerialName("vendorListId") val vendorListId: String?,
     @SerialName("webConsentPayload") val webConsentPayload: JsonObject? = null,
+    @SerialName("expirationDate") var expirationDate: String?,
```
bohdan-go-wombat, line 93: thanks for the comma at the end 👍

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/campaign/CampaignManagerTest.kt
```diff
@@ -199,6 +199,7 @@ class CampaignManagerTest {
             status = CcpaStatus.rejectedSome,
             uuid = "fake_uuid",
             webConsentPayload = JsonObject(mapOf()),
+            expirationDate = null
```
bohdan-go-wombat, line 202: can we add comma at the end, for the future convenience?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/util/SpUtilsTest.kt
```diff
@@ -142,6 +142,7 @@ internal class SpUtilsTest {
         status = ccpaStatus,
         type = CampaignType.CCPA,
         url = null,
-        webConsentPayload = null
+        webConsentPayload = null,
+        expirationDate = null
```
bohdan-go-wombat, line 146: can we add comma at the end, for the future convenience?

samples/app/src/androidTest/java/com/sourcepoint/app/v6/TestData.kt
```diff
@@ -311,4 +282,171 @@ object TestData {
       }
     """.trimIndent()
 
+    val storedConsentV742 = """
```
bohdan-go-wombat, line 285: I think that would be better if we update `storedConsentV741` in order to see if consent of v7.4.2 is compatible with our tests

bohdan-go-wombat (commented):
have a couple of questions and suggestions here

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/campaign/CampaignManagerImplTest.kt
carmelo-iriti: the test is: GIVEN_an_expired_GDPR_and_a_valid_CCPA_campaign_DELETE_only_the_GDPR_consent
this mean that there shouldn’t be keys belonging to the GDPR campaign

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/campaign/CampaignManagerImplTest.kt
carmelo-iriti: same logic as before but different campaign

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: Exactly if the format changes and we don’t get an exception we can have a false positive

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: like before

samples/app/src/androidTest/java/com/sourcepoint/app/v6/TestData.kt
carmelo-iriti: v7.4.1 cannot be changed anymore let’s keep it like this

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/Utils.kt
```diff
@@ -0,0 +1,32 @@
...
+                check { it.second as? Int }?.let { v -> spEditor.putInt(it.first, v) }
+            }
+            spEditor.apply()
+        }
+
+        fun Context.spEntries(): Map<String, Any?> {
+            return PreferenceManager.getDefaultSharedPreferences(this).all
+        }
+    }
+}
```
andresilveirah, line 32: @carmelo-iriti does this code needs to be in the SDK? Seems like it's only used for tests, right?

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpaImplTest.kt
andresilveirah: @carmelo-iriti was this addressed?

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdprImplTest.kt
andresilveirah: @carmelo-iriti was this addressed?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
andresilveirah: @carmelo-iriti so what's the solution? if `formatter.parse(it)` throws an exception, the gdpr consent will never expire?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
andresilveirah: @carmelo-iriti so what's the solution? if formatter.parse(it) throws an exception, the ccpa consent will never expire?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageCcpa.kt
andresilveirah: @carmelo-iriti was Bohdan's comment addressed? why do we need to store this key in the first place?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdpr.kt
andresilveirah: Can expiration date be part of the consent object we store in the local storage rather than a separate key?

cmplibrary/src/androidTest/java/com/sourcepoint/cmplibrary/data/local/DataStorageGdprImplTest.kt
carmelo-iriti: @andresilveirah we can decide together with @bohdan-go-wombat how to rename all the tests in a way that makes everyone happy

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: There are 2 cases that we can have a crash:
- the value to parse is null but we are already have a null check
- the value from the backend changes and we fallback to the graceful degradation.
If something changes in the backend, we'll know from our Datadog metrics before next year when the consent will expire

PR #735, opened 2023-11-07
[DIA-2786] Added support fo the new CCPA OTT PM 

carmelo-iriti:


README.md
```diff
@@ -306,13 +338,16 @@ An OTT privacy manager can be **resurfaced** for your project (_e.g. via a butto
 Kotlin
 
 ```kotlin
-    spConsentLib.loadPrivacyManager("<PM_ID>", CampaignType.GDPR) // For a GDPR campaign
+
+import java.awt.TrayIcon.MessageType
```
andresilveirah, line 342: Maybe this `import` statement is incorrect?

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/SpConsentLibImpl.kt
```diff
@@ -643,7 +696,7 @@ internal class SpConsentLibImpl(
                                 env = env,
                                 campaignType = actionImpl.campaignType,
                                 pmConfig = pmUrlConfig,
-                                messSubCat = TCFv2,
+                                messageType = context.toMessageType(),
```
andresilveirah, line 699: Strange that the `messSubCat` was `TCFv2` even for CCPA.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/util/HttpUrlManager.kt
```diff
@@ -119,19 +119,21 @@ internal object HttpUrlManagerSingleton : HttpUrlManager {
...
+            MOBILE -> "ccpa_pm/index.html"
+        }
 
         return HttpUrl.Builder()
             .scheme("https")
             .host(env.pmHostCcpa)
             .addPathSegments(pathSegment)
             .addQueryParameter("site_id", pmConf.siteId)
             .addQueryParameter("preload_consent", true.toString())
+            .addQueryParameter("is_ccpa", true.toString())
```
andresilveirah, line 136: Is this query param necessary?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SpConsentLibImplTest.kt
```diff
@@ -121,15 +122,14 @@ class SpConsentLibImplTest {
             )
         }
 
-        verify { logger.i(any(), "${OTT.code}") }
+        verify { logger.i(any(), LEGACY_OTT.name) }
```
andresilveirah, line 125: Are you testing the `logger`?

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/data/network/util/HttpUrlManagerTest.kt
```diff
@@ -197,8 +317,8 @@ class HttpUrlManagerTest {
             messageId = "111",
             siteId = null
         )
-        val sut = HttpUrlManagerSingleton.pmUrl(Env.PROD, CampaignType.CCPA, config, messSubCat = TCFv2).toString()
-        sut.assertEquals("https://cdn.privacy-mgmt.com/ccpa_pm/index.html?site_id&preload_consent=true&ccpaUUID=uuid&message_id=111&scriptType=android&scriptVersion=${BuildConfig.VERSION_NAME}")
+        val sut = HttpUrlManagerSingleton.pmUrl(Env.PROD, CampaignType.CCPA, config, messageType = LEGACY_OTT,).toString()
+        sut.assertEquals("https://cdn.privacy-mgmt.com/ccpa_ott/index.html?site_id&preload_consent=true&is_ccpa=true&ccpaUUID=uuid&message_id=111&scriptType=android&scriptVersion=${BuildConfig.VERSION_NAME}")
```
andresilveirah, line 321: Same here, not sure if `is_ccpa` is required.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/util/HttpUrlManager.kt
carmelo-iriti: @mguerra10128 wrote in a comment in the ticket to add this query param for the ccpa

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SpConsentLibImplTest.kt
carmelo-iriti: this test check that whenever the user select a MessageType from the public api `loadPrivacyManager`, the SDK is generating the correct url type for the pm

carmelo-iriti: the logger is a spy object

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/util/HttpUrlManager.kt
andresilveirah: Thanks @carmelo-iriti , just checked the rendering app's code base and it seems to be used by the OTT messages https://github.com/search?q=repo%3ASourcePointUSA%2Fmessage-v2-preact%20isCcpa&type=code

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SpConsentLibImplTest.kt
andresilveirah: Not sure if I understood you well. If we remove the log from `loadPrivacyManager` method, will the test fail?

carmelo-iriti: yes

andresilveirah: I think we need to move away from using the loggers as spy objects. It's not a robust solution in my opinion, because it relies on a side effect (the loggers being called). 

In this specific case we can:
* extract the logic smaller functions and test those via unit tests
* spy on the actual webview object to make sure the right URL was passed to load url via unit/integration tests 
* assert the right message renders via UI tests.

Since the SDK is already relying on loggers for testing, we should address that in a different PR.

andresilveirah (approved):

cmplibrary/src/test/java/com/sourcepoint/cmplibrary/SpConsentLibImplTest.kt
carmelo-iriti: I see what you mean, this test here is an integration test that runs locally without the need of instancing an Emulator (which requires to build a test apk and then start an emulator and this requires at least 30-40 sec).
This test itself runs in 2 sec instead of a UI or Instrumentation test that requires 15 sec

PR #740, opened 2023-11-09
[DIA-2836] UI test for the GDPR applies fix

bohdan-go-wombat:


samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinTest.kt
```diff
@@ -1155,6 +1155,42 @@ class MainActivityKotlinTest {
         }
     }
 
+    /**
+     * UI test that verifies the change of the GDPR applies value after the user moves out from the
+     * place where GDPR does not apply and moves to the place where GDPR applies.
+     */
+    @Test
+    fun given_the_user_changes_location_then_should_update_applies_value() = runBlocking<Unit> {
```
carmelo-iriti, line 1163: please use this convention 
GIVEN_initial_condition_SHOULD_expected_result

samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinTest.kt
```diff
@@ -113,14 +114,13 @@ class MainActivityKotlinTest {
         +(CampaignType.CCPA)
     }
 
-    private val spConfBohdan = config {
+    private val spConfigMobileBohdanTestGdpr = config {
```
carmelo-iriti, line 117: please don't use personal name 😅

samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinTest.kt
```diff
@@ -1187,16 +1187,15 @@ class MainActivityKotlinTest {
...
         scenario = launchActivity()
 
-        wr(backup = { clickOnRefreshBtnActivity() }) { tapAcceptAllOnWebView() }
-        wr { tapAcceptAllOnWebView() }
-
         wr { verify(exactly = 0) { spClient.onError(any()) } }
         wr { verify(exactly = 1) { spClient.onSpFinished(any()) } }
 
         // TODO check if the applies value of GDPR changed to TRUE
         scenario.onActivity { activity ->
```
carmelo-iriti, line 1194: you need to wrap this condition into a wr{} otherwise the test will skip it

samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinTest.kt
bohdan-go-wombat: I know, but the name of the property has my name in it :D

carmelo-iriti: I saw it.. Creating a property requires time so I didn’t feel to ask it

samples/app/src/androidTest/java/com/sourcepoint/app/v6/MainActivityKotlinTest.kt
bohdan-go-wombat: > you need to wrap this condition into a wr{} otherwise the test will skip it

why would the test skip it?

I thought that `wr { ... }` is a wrapper that checks whether or not the code block inside it fails (there is a `try/catch` block inside), we also have plenty of tests that are not using that wrapper and they all seems to work fine

carmelo-iriti: right usually there are assertions that check some parameters

bohdan-go-wombat: @carmelo-iriti so if there is a check, we don't need a `wr { ... }` wrapping those lines, right?

carmelo-iriti (approved):

PR #741, opened 2023-11-10
[DIA-2620] Update /consent-status to support usnat

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -73,17 +73,20 @@ internal interface CampaignManager {
     // Consent Status
     var gdprConsentStatus: GdprCS?
     var ccpaConsentStatus: CcpaCS?
+    var usNatConsentStatus: USNatConsentData?
```
bohdan-go-wombat, line 76: I think it is better to rename it to `usNatConsentData`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -567,6 +583,10 @@ private class CampaignManagerImpl(
             dataStorage.ccpaConsentUuid = value
         }
 
+    override var usNatUuid: String?
+        get() = usNatConsentStatus?.uuid
+        set(value) { usNatConsentStatus?.uuid = value }
```
bohdan-go-wombat, line 588: I'm not sure if that way you will achieve what you want to, I don't think this line will actually update the `uuid` of the `usNatConsentStatus` from the `DataStorage`. Meaning that, if we set `usNatUuid` and then call `usNatConsentStatus?.uuid` we will get different results.

I think it should be something like:
```
set(value) {
    usNatConsentStatus = usNatConsentStatus.copy(uuid = value)
}
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -665,6 +685,12 @@ private class CampaignManagerImpl(
             dataStorage.ccpaDateCreated = value
         }
 
+    override var usNatDateCreated: String?
+        get() = usNatConsentStatus?.dateCreated
+        set(value) {
+            usNatConsentStatus?.dateCreated = value
```
bohdan-go-wombat, line 691: same here

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/local/DataStorageUSNat.kt
```diff
@@ -0,0 +1,49 @@
...
+    context: Context
+): DataStorageUSNat = DataStorageUSNatImpl(context)
+
+private class DataStorageUSNatImpl(context: Context) : DataStorageUSNat {
+
+    override val preference: SharedPreferences by lazy {
+        PreferenceManager.getDefaultSharedPreferences(context)
+    }
+
+    override var usNatConsentStatus: String?
```
bohdan-go-wombat, line 31: I think we should rename it to `usNatConsentData`, since the signature of the object is `UsNatConsentData`, and we should keep the naming convention straight, from my pov

bohdan-go-wombat (commented):
left a couple of comments

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: the class with lower case?

bohdan-go-wombat: no, the name of the param itself:
```
var usNatConsentData: USNatConsentData?
```

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: changing like this
`
override val usNatUuid: String?
        get() = usNatConsentStatus?.uuid
`

carmelo-iriti: I'll just delete this property

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
carmelo-iriti: same deletion

bohdan-go-wombat (approved):

PR #743, opened 2023-11-13
[DIA-2619] Update `POST /choice` to support UsNat

bohdan-go-wombat:


carmelo-iriti:
looks good

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/PostChoiceApiModelExt.kt
```diff
@@ -72,3 +72,38 @@ internal fun postChoiceCcpaBody(
         }
     }
 }
+
+internal fun postChoiceUsNatBody(
+    granularStatus: ConsentStatus.GranularStatus?,
+    idfaStatus: String? = null,
```
carmelo-iriti, line 78: this is only for ios, you can remove it

carmelo-iriti:
it looks good, we need just to fix the tests

carmelo-iriti (approved):

PR #744, opened 2023-11-14
[DIA-2618] Update /pv-data to support usnat

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -478,8 +479,7 @@ private class CampaignManagerImpl(
                 isGdprPresent[$isGdprPresent] - isCcpaUuidPresent[$isCcpaUuidPresent]
                 isUSNatUuidPresent[$isUsNatUuidPresent] - isLocalStateEmpty[$isLocalStateEmpty]
                 isV6LocalStatePresent[$isV6LocalStatePresent] - isV6LocalStatePresent2[$isV6LocalStatePresent2]
-                hasNonEligibleLocalDataVersion[$hasNonEligibleLocalDataVersion]
-                isEligibleForCallingConsentStatus[$isEligibleForCallingConsentStatus]
+                hasDataVersion[$hasNonEligibleLocalDataVersion] - shouldCallConsentStatus[$isEligibleForCallingConsentStatus]
```
carmelo-iriti, line 482: last line was not readable from metaapp

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/util/SpUtils.kt
```diff
@@ -121,3 +118,6 @@ internal fun userConsents(
             }
     )
 }
+
+internal fun SpConfig.isIncluded(campaign: CampaignType) =
```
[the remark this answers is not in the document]
bohdan-go-wombat: Actually, from my pov, there are 2 ways we can store those ext functions:
1. store them by type in each file accordingly (meaning that for example if we create extension function for `SpConfig` class, then we create file `SpConfigExt` and add all the ext functions there)
2. store them by logical affiliation (meaning that we have ext functions that are being used only in a certain flow/layer)

Since we don't have a very strict layer separation, I think the first way is the best one for us

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -259,7 +270,7 @@ private class ServiceImpl(
                 execManager.executeOnMain { showConsent() }
             }
 
-            val isGdprInConfig = spConfig.campaigns.find { it.campaignType == GDPR } != null
+            val isGdprInConfig = spConfig.isIncluded(GDPR)
```
bohdan-go-wombat, line 273: kudos for that change 👍

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/consent/ConsentManagerUtils.kt
```diff
@@ -168,9 +167,26 @@ private class ConsentManagerUtilsImpl(
             }
         }
 
-    override var messagesResp: MessagesResp?
```
bohdan-go-wombat, line 171: thanks for clearing this out 👍

bohdan-go-wombat (commented):
left a couple of comments

PR #757, opened 2023-11-30
[DIA-2957] Expire consent based on `vendorListId` from `/meta-data` response

bohdan-go-wombat:
JIRA - [DIA-2957](https://sourcepoint.atlassian.net/browse/DIA-2957)

[DIA-2957]: https://sourcepoint.atlassian.net/browse/DIA-2957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
@@ -101,7 +101,7 @@ internal interface CampaignManager {
     fun getCcpaPvDataBody(messageReq: MessagesParamReq): JsonObject
     fun getUsNatPvDataBody(messageReq: MessagesParamReq): JsonObject
     fun deleteExpiredConsents()
-    fun clearConsentIfVendorListIdChanged(gdprVendorListId: String?,usNatVendorListId: String?)
+    fun clearConsentIfVendorListIdChanged(gdprVendorListId: String?, usNatVendorListId: String?)
 
     companion object {
         const val SIMPLE_DATE_FORMAT_PATTERN = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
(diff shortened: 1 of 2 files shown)


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
```diff
@@ -697,6 +704,49 @@ private class CampaignManagerImpl(
         }
     }
 
+    /**
+     * Method that clears out local consent if the vendor list id of a specific campaign changed
+     */
+    private fun clearConsentIfVendorListIdChanged(
```
carmelo-iriti, line 710: This method is not testable.
Just create a property that return a boolean value (like CamapaignManager:85) called for instance `hasGdpr(Usnat)VendorListChanged` so we can test the logic and then we can use the same structure of `deleteExpiredConsents` CampaignManager:936

carmelo-iriti (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/campaign/CampaignManager.kt
bohdan-go-wombat: @carmelo-iriti we are not able to create a **property**, we have to create a method that will receive a vendor list id parameter from the response of `/meta-data` call and compare it to the stored one (as we already have - `isNewGdprVendorListId(gdprVendorListId)` and `isNewUsNatVendorListId(usNatVendorListId)`, maybe we can actually rename those so that the naming is more descriptive)

and I don't see how this method is _"untestable"_;
1. for unit tests we can actually mock stored `MetaDataResp` and pass different `gdprVendorListId` and `usNatVendorListId` to the function and see if the proper methods from the `dataStorage` was called
2. for the UI tests, we can actually use stored consent and then run a test with a property that has different `vendorListId` and check if the consent was cleared out

let me know if that makes senes

ciriti: - This method is `private` how do you test a private method?
- if you cannot create a property, you can turn it into a function `hasGdpr(Usnat)VendorListChanged(para)` that return a boolean, after you can create a unit test for it

carmelo-iriti: I used the other account by mistake

bohdan-go-wombat: @carmelo-iriti I will do the following:
1. I will change modifier to `internal`
2. we already have those methods, we just have to rename those to be more descriptive

PR #760, opened 2023-12-04
[DIA-3052] Add re-consent step to usnat

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/MetaDataApiModel.kt
```diff
@@ -45,6 +45,7 @@ data class MetaDataResp(
     data class USNat(
         @SerialName("_id") val vendorListId: String?,
         @SerialName("additionsChangeDate") val additionsChangeDate: String?,
+        @SerialName("legalBasisChangeDate") val legalBasisChangeDate: String?,
```
andresilveirah, line 48: @carmelo-iriti same here, we should remove `legalBAsisChangeDate`.

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/MessagesApiModel.kt
```diff
@@ -156,6 +156,8 @@ data class USNatConsentStatus(
     @SerialName("consentedToAny") val consentedToAny: Boolean?,
     @SerialName("granularStatus") val granularStatus: USNatGranularStatus?,
     @SerialName("hasConsentData") val hasConsentData: Boolean?,
+    @SerialName("legalBasisChanges") var legalBasisChanges: Boolean? = null,
```
andresilveirah, line 159: @carmelo-iriti I think it's safe to remove legalBasisChangeDate for now since it won't be used on usnat.

andresilveirah (changes requested):

PR #761, opened 2023-12-04
[DIA-2957] Expire consent based on `vendorListId` from `/meta-data` response

bohdan-go-wombat:
JIRA - [DIA-2957](https://sourcepoint.atlassian.net/browse/DIA-2957)

[DIA-2957]: https://sourcepoint.atlassian.net/browse/DIA-2957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Nevazhnovu (approved):

carmelo-iriti:
It seems that the test, which need a fix, doesn't actually do anything, can you please verify it?

bohdan-go-wombat:
@carmelo-iriti the test that you are talking about is `GIVEN_the_user_changes_location_SHOULD_update_gdpr_applies_value`, and it actually does verify the change of applies value of the GDPR campaign, because in the stored data it is true:
```
"sp.gdpr.key.applies": true,
```
and then we verify if the value changed:
```
getBoolean("sp.gdpr.key.applies", true).assertFalse()
```

carmelo-iriti:
I see.. that's not the point, what I meant is that test is a false positive, it never test the condition

carmelo-iriti:
I'm going to create a ticket for it, we can merge this story for now

PR #762, opened 2023-12-05
[DIA-3213] Consentstring as an array

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -82,7 +82,14 @@ data class USNatConsentData(
     @SerialName("type") override val type: CampaignType = CampaignType.USNAT,
     @SerialName("url") override val url: String?,
     @SerialName("expirationDate") override val expirationDate: String?
-) : CampaignMessage
+) : CampaignMessage {
+    @Serializable
+    data class ConsentStrings(
```
andresilveirah, line 87: @carmelo-iriti I believe the correct name for this data class should be `ConsentString` (singular). Then `USNatConsentData` will contain an attribute called `consentStrings`, which is a list of `ConsentString` .

andresilveirah (changes requested):

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
carmelo-iriti: sounds fair

PR #764, opened 2023-12-06
[DIA-3227] Handle GPP data for usnat

carmelo-iriti:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/JsonToMapExt.kt
```diff
@@ -34,7 +34,7 @@ internal fun Map<String, Any?>.toJSONObj(): JSONObject {
     return JSONObject(this)
 }
 
-internal fun Map<String, Any?>.toTcfJSONObj(): JSONObject {
+internal fun Map<String, Any?>.toConsentJSONObj(): JSONObject {
```
andresilveirah, line 37: Nice! Reusing the same mechanism for both TCF and GPP data.

andresilveirah (approved):

PR #806, opened 2024-04-15
[HCD 502] Add USP String support for MSPS

tom-win87:


tom-win87:
Please review the code examples for both kotlin and java in the added section

PR #811, opened 2024-04-29
Update test dependencies

andresilveirah:


.github/workflows/sample_app_tests_local.yml
```diff
@@ -0,0 +1,52 @@
+name: Sample App UI Tests (Local)
```
bohdan-go-wombat, line 1: wdym by local? this is a configuration for UI tests on GitHub actions, so those are not `local`, those are `remote` tests

andresilveirah: @bohdan-go-wombat by local I meant "running on GH instead of Firebase", I wanted to check if we could get rid of firebase (and we can!)

PR #823, opened 2024-06-05
[HCD 525] - Clearing authId

tom-win87:


andresilveirah:
Thank you @tom-win87 !

PR #837, opened 2024-10-16
DIA-4613 replace native `/consent-status` with core-module's 

andresilveirah:


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -36,45 +34,48 @@ enum class GranularState {
...
+    @SerialName("rejectedAll") val rejectedAll: Boolean? = null,
+    @SerialName("rejectedCategories") val rejectedCategories: List<String>? = emptyList(),
+    @SerialName("rejectedVendors") val rejectedVendors: List<String>? = emptyList(),
+    @SerialName("signedLspa") val signedLspa: Boolean? = null,
     @SerialName("uspstring") val uspstring: String? = null,
-    @Serializable(with = CcpaStatusSerializer::class) val status: CcpaStatus?,
-    @SerialName("GPPData") @Serializable(with = JsonMapSerializer::class) val gppData: Map<String, JsonElement>? = null,
-    @SerialName("uuid") var uuid: String?,
+    @Serializable(with = CcpaStatusSerializer::class) val status: CcpaStatus? = null,
+    @SerialName("GPPData") @Serializable(with = JsonMapSerializer::class) val gppData: Map<String, JsonElement>? = emptyMap(),
```
Nevazhnovu, line 57: If there a particular reason why we now initialising all these values? 
Main reason for me to ask is: few variables might become NOT nullable if they are already initialised ( `gppData`, `rejectedCategories`, `rejectedVendors`)

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -85,13 +86,49 @@ data class USNatConsentData(
     @SerialName("uuid") var uuid: String? = null,
     @SerialName("webConsentPayload") val webConsentPayload: JsonObject? = null,
     @SerialName("message") override val message: JsonElement? = null,
-    @SerialName("GPPData") @Serializable(with = JsonMapSerializer::class) val gppData: Map<String, JsonElement>? = null,
+    @SerialName("GPPData") @Serializable(with = JsonMapSerializer::class) val gppData: Map<String, JsonElement>? = emptyMap(),
```
Nevazhnovu, line 89: The same concern as above (about initialising nullable variable)

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
```diff
@@ -113,31 +150,63 @@ data class USNatConsentData(
...
+    @SerialName("specialFeatures") val specialFeatures: List<String>? = emptyList(),
+    @SerialName("vendors") val vendors: List<String>? = emptyList(),
+    @SerialName("addtlConsent") val addtlConsent: String? = null,
+    @SerialName("consentStatus") val consentStatus: ConsentStatus? = null,
+    @SerialName("cookieExpirationDays") val cookieExpirationDays: Int? = null,
+    @SerialName("customVendorsResponse") val customVendorsResponse: CustomVendorsResponse? = null,
+    @SerialName("dateCreated") val dateCreated: String? = null,
+    @SerialName("euconsent") val euconsent: String? = null,
+    @Serializable(with = GrantsSerializer::class) val grants: Map<String, GDPRPurposeGrants>? = emptyMap(),
+    @Serializable(with = JsonMapSerializer::class) val TCData: Map<String, JsonElement>? = emptyMap(),
```
Nevazhnovu, line 170: The same concern as above. In this class it's regarding `categories`, `legIntCategories`, `legIntVendors` , `specialFeatures` , `vendors` , `grants` , `TCData`

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/ServiceImpl.kt
```diff
@@ -725,30 +729,51 @@ internal class ServiceImpl(
...
+                uuid = usNatConsentData?.uuid,
+                hasLocalData = false, // usNatConsentData != null && usNatConsentData?.uuid != null,
+                transitionCCPAAuth = transitionCCPAAuth,
+                optedOut = transitionCCPAOptedOut
+            ) },
+            ccpa = campaigns4Config.firstOrNull { it.campaignType == CCPA }?.let { ConsentStatusRequest.MetaData.Campaign(
+                applies = ccpaApplies ?: false,
+                dateCreated = ccpaConsentStatus?.dateCreated,
+                uuid = ccpaConsentStatus?.uuid,
+                hasLocalData = false, // ccpaConsentStatus != null && ccpaConsentStatus?.uuid != null
```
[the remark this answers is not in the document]
andresilveirah: Good point. I should remove the comments. On usnat, if `hasLocalData` is true, the endpoint doesn't return consent data and the SDK overwrites the local data with empty/null values. That's why we're hardcoding false here (we probably could do that at the core-module level).

cmplibrary/src/main/java/com/sourcepoint/cmplibrary/data/network/model/optimized/ConsentStatusApiModel.kt
andresilveirah: Attributes that are a collection, it's a better practice to give them a default value (like empty collection) instead of a null value. The only reason I kept them as optionals (`?`) is because they might not be present in the responses we get further along the code. We should address that in the mobile-core in the future.

PR #853, opened 2025-04-08
Expand list of supported message languages

andresilveirah:
* Added support to the following languages:
  * sq
  * zh-hant
  * ka
  * hi
  * mt
  * pt-br
  * pt-pt
  * sw
  * th
  * uk
  * vi
  * cy
* changed language short code from upper to lower case
* updated `mobile-core`


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/model/MessageLanguage.kt
```diff
@@ -3,48 +3,57 @@ package com.sourcepoint.cmplibrary.model
...
+    SERBIAN_LATIN("sr-latn"),
+    SLOVAK("sk"),
+    SLOVENIAN("sl"),
+    SPANISH("es"),
+    SWAHILI("sw"),
+    SWEDISH("sv"),
+    TAGALOG("tl"),
+    THAI("th"),
+    TURKISH("tr"),
+    UKRAINIAN("uk"),
```
Nevazhnovu, line 54: Let's gooooo 🗣🗣🗣

Nevazhnovu (approved):

PR #856, opened 2025-04-18
[DIA-5502] Fix AndroidTv navigation 

Nevazhnovu:
[DIA-5502](https://sourcepoint.atlassian.net/browse/DIA-5502) [AndroidTv] Navigation Broken on 7.12.0

- [x] Use `super.dispatchKeyEvent` in case `event.keyCode` is not related to the back button

[DIA-5502]: https://sourcepoint.atlassian.net/browse/DIA-5502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/mobile_core/SPConsentWebView.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/mobile_core/SPConsentWebView.kt
@@ -171,6 +171,8 @@ class SPConsentWebView(
                     )
                 )
             }
+        } else {
+            super.dispatchKeyEvent(event)
         }
         return true
     }
--- a/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/mobile_core/SPConsentWebView.kt
+++ b/cmplibrary/src/main/java/com/sourcepoint/cmplibrary/mobile_core/SPConsentWebView.kt
@@ -171,10 +171,10 @@ class SPConsentWebView(
                     )
                 )
             }
+            return true
         } else {
-            super.dispatchKeyEvent(event)
+            return super.dispatchKeyEvent(event)
         }
-        return true
     }
 
     override fun load(


cmplibrary/src/main/java/com/sourcepoint/cmplibrary/mobile_core/SPConsentWebView.kt
```diff
@@ -171,6 +171,8 @@ class SPConsentWebView(
                     )
                 )
             }
+        } else {
+            super.dispatchKeyEvent(event)
```
andresilveirah, line 175: @Nevazhnovu maybe we should change the code to be:
```kotlin
override fun dispatchKeyEvent(event: KeyEvent): Boolean {
        if (event.keyCode == KeyEvent.KEYCODE_BACK && event.action == KeyEvent.ACTION_DOWN) {
            if (messageType == OTT || messageType == LEGACY_OTT) {
                evaluateJavascript("window.postMessage({name:\"sp.BACK\"})", null)
            } else {
                onAction(
                    view = this,
                    action = SPConsentAction(
                        actionType = if (isFirstLayer) ActionType.MSG_CANCEL else ActionType.PM_DISMISS,
                        campaignType = campaignType,
                        messageId = ""
                    )
                )
            }
            return true
        } else {
            return super.dispatchKeyEvent(event)
        }
}

andresilveirah:
@Nevazhnovu did you have the chance to manually test the navigation on AndroidTV? was it really just the back button not working? Is there any other key code we could use instead of `KeyEvent.KEYCODE_BACK` ?

Nevazhnovu:
> @Nevazhnovu did you have the chance to manually test the navigation on AndroidTV? was it really just the back button not working? Is there any other key code we could use instead of `KeyEvent.KEYCODE_BACK` ?

@andresilveirah everything was not working, except the back button. Back button was working fine. Now everything works 💪 I did test it with android TV emulator

PR #861, opened 2025-06-10
Add `GlobalCmpConsent` to `SPConsents`

Nevazhnovu:


andresilveirah (approved):
It's approved. Let's wait for global cmp to be in prod before merging it

samples/app/src/main/java/com/sourcepoint/app/v6/di/AppModule.kt
```diff
@@ -30,13 +30,17 @@ val appModule = module {
     single<DataProvider> {
         val gdprPmId = "488393" // stage "13111"
         val ccpaPmId = "509688" // "14967"
+        val usnatPmId = "943886"
```
andresilveirah, line 33: @Nevazhnovu the correct PM id should be `988851`

